Issue #16, opened 2016-11-25
Firmware update procedure requires to completely wipe SD card


jens-maus:
Currently, upon a new RaspberryMatic release all users still have to completely wipe their SD card, update it with the next image and then restore a backup. This is quite cumbersome and error prone. For proper updates in future we should support normal "firmware updates" like this is currently possible for the old CCU2 hardware. For this, we have to lookup how this is currently done and how we can automatically generate similar "firmware update" archives users can then directly upload via the WebUI.

j-a-n:
I began to analyze the original firmware update process on the CCU2:

1.) firmware update archive (.tgz) is uploaded via /www/config/cp_maintenance.cgi to /tmp/
2.) the files "update_script", "EULA.en", "EULA.de" and "EULA.tr" are extracted from the uploaded archive to /var directory
3.) update process aborts if update_script is not present in archive
4.) firmware update archive is moved from /tmp to /var/new_firmware.tar.gz
5.) EULA is displayed and has to be accepted by user
6.) If user accepts EULA the system reboots
7.) the new_firmware.tar.gz seems to get extracted at boot and the update_script is executed
original update_script:
`#!/bin/sh

STARTPWD=`pwd`
ROOT_FS=rootfs.ubi
KERNEL_IMAGE=uImage

ROOT_MTDCHAR=/dev/mtd7

KERNEL_MTDCHAR=/dev/mtd5

if [ -f $STARTPWD/$KERNEL_IMAGE ]; then
 	echo "Erasing kernel partition ..."            
 	flash_erase $KERNEL_MTDCHAR 0 0                      
 	echo "Writing kernel to NAND ..."         
  	nandwrite -p $KERNEL_MTDCHAR $KERNEL_IMAGE
fi

if [ -f $STARTPWD/$ROOT_FS ]; then
  	echo "Writing rootfs to NAND ..."
	ubidetach -p $ROOT_MTDCHAR
	ubiformat $ROOT_MTDCHAR -f $ROOT_FS
fi
`

Changes needed for RaspberryMatic:
1.) cp_maintenance.cgi uses tar to extract files from the firmware update archive.
The tar version included in RaspberryMatic needs different parameter order:
original (not working): `exec tar zxvf [lindex $firmware_file 0] update_script EULA.en EULA.de EULA.tr -C /var/`
working command: `exec tar zxvf [lindex $firmware_file 0] -C /var update_script EULA.en EULA.de EULA.tr`

[...]

PR #123, opened 2017-06-28
firmware upgrade feature￼

j-a-n:
I implemented the update process as described in #16.
It is now possible to upgrade to the latest available version via webui.
Additionally I implemented the commandline script _firmware-tool_ which provides other options, i.e. installing a specific firmware version.
We will have to discuss the right size for the root filesystem, which I increased to 800MB.
I appreciate any feedback.


jens-maus:
First of all, thank you very much for your efforts and the PR. I really appreciate your help.

However, as you might have seen, I also already started to integrate the webui-based update/upgrade feature during the last days and unfortunately your PR seem to conflict with my changes as your branch isn't based on the latest changes in the webui-upgrade-feature branch. So please at least rebase your changes on the latest changes in that branch.

Furthermore, as I have quickly reviewed your PR, I think we should first talk about the principle approach how to integrate such a feature in RaspberryMatic because your approach seems to be slightly different to what I have in mind. As you might have seen I have taken a slightly different road to integrate the two-root-partition approach of providing a webui-based upgrade mechanism. So my ideas are as follows:

1. Keep the root partition 500MB max. (it is currently filled by 74% only and I don't think we will increase the content of the rootfs by more than 120MB anytime soon). In addition, if this really happens we can then think about repartitioning the rootfs on-the-fly in some manner. However, just increasing it now doesn't seem to be necessary IMHO.
2. As I see, your script is automatically downloading updates from GitHub without requiring the user to upload the update himself. However, I want to keep the webui-based update mechanism as compatible as possible to the way the CCU2 update functionality works. That's the reason why I have invested some time already in patching `cp_maintenance.cgi` to directly allow `*.zip` file uploads to `/usr/local/tmp`, unarchiving it and even checking its sha256 checksum before asking the user to confirm the update process.
3. The way you are currently performing the update/upgrade is by running the upgrade while the normal system is still running in the background. For security reasons and to keep the upgrade process compatible to the CCU2 way I want to implement the upgrade to be actually executed when the RaspberryMatic system is shutting down (e.g. initiated when `S00eQ3SystemStarted` is running its 'stop' routine) with all normal services being shut down when the lofs mount happens, the rsync starts, etc. And my current plans are to even perform this using a simple shell-script rather than using tcl if this is fine with you.
4. Your approach doesn't consider keeping a backup for the `/boot` filesystem in any way. Please keep in mind that newer RaspberryMatic version will come with newer kernels and raspberry firmware files and currently you are simply overwriting the old ones. But I think we should think about keeping a backup of the main files just in case the update doesn't work and the user wants to revert to the old version on the other rootfs partition.

I know that all this sounds like we would have to significantly rework your update mechanism and even throw away things not required (automatic GitHub updates, etc.), but I still hope we can work together to get the webui-based update mechanism integrated in the most smooth and CCU2 compatible way. Looking forward to hearing your opinion.

j-a-n:
1.) If it ever happens, repartitioning the running system would be nearly impossible.
2.) I prefer my approach because its much easier.
3.) I don't see the advantage here.
4.) That is right, but I think if the system is no longer able to boot, it is also not possible to revert the changes. In that case you would have to reinstall.

I think you have a clear idea of how you want things to be, which makes it hard for me to argue.
Don't get me wrong but I will stop investing my time into this integration now.
For me this project is not too important and I am confident that your approach will be good to.

jens-maus:
Well, thanks for your understanding. I really want to make sure you are not frustrated and I want to point out that I still want you onboard in integrating this long-open feature request. And I really appreciate all the time you invested in preparing integration of your approach. Nevertheless, I still have to keep compatibility to CCU2 in some minor way and thus I don't want to introduce a completely different way of performing the upgrade (from end-user perspective). And regarding the single points I still hope we can continue our discussion because I really want to use your expertise since you have invested quite some time into this topic:

1. I still think repartitioning should be possible, especially if the update/upgrade is performed upon shutting down the system, because then we can unmount /usr/local and resize it, etc. However, I agree that this might require some additional work and thoughts – still I believe that we won't pass the 500MB anytime soon. And if so, users could still easily be advised to start from a new sd card for the time being.

2. It is much easier for you, I guess :) From the end user perspective I think users are used to download the update files themselves and then upload them to their CCU manually. This will also ensure that, e.g. after a potential change of the download URL, etc. the update process still works.

3. The advantage is that by having shut down the system we should be more on the safe if something crashes the system. In addition, we could then also umount /usr/local and repartition it in the future or do a file system check, etc.

4. If the system is not bootable anymore we could at least give users the advice to mount the sd card manually on a PC/Mac and then change some files so that the system starts booting form the old rootfs partition again.

So for my proposed changes. I will try to integrate my procedure during the next days so that you can hopefully test it yourself and give some feedback or further suggestions. Is that fine for you?

j-a-n:
1.) Believe me, repartitoning the running system will get very complicated. I asked you before - why is the size of rootfs so important for you? If image size is the reason, why not resize rootfs1 during checkFactoryReset() and then stick with it? This would result in an even smaller image size than today.

2.) I think end users would love to switch their habits in that case.

3.) I disagree. If you upgrade while the system is running, you can easily inform the user about any problem. If update errors occur on shutdown, the system would just hang or reboot.

4.) If you think this is helpful it would be easy to backup the files from /boot. Perhaps it would be sufficient to just backup/rename the kernel image file.

jens-maus:
1. How would you do that? the rootfs is already mounted at the time S00eQ3SystemStart is executed, thus this would require to resize the rootfs online while it is mounted. Probably not a very good idea
2. I am not sure. Also keep in mind that I need to stay a bit compatible to the way/webui how this is done on a CCU/OCCU also in terms of support, etc.
3. You are right that an update progress would better be shown while the system is running, however, we can also put a logfile somewhere with debug information. And then we can even show update progress on the HDMI console for the same debugging reasons.
4. That's exactly what I have planned ;)

j-a-n:
1.) Not a problem if you do not move the partition start point. ext4 can be easily and safely resized online (mounted). You could even delete the partition and recreate it in MBR as long as you do not change the starting point.

einfall:
Any news on this Feature ?

jens-maus:
I am still working on it and even integrating a rescue type system at the same time. So please some more patience ;)

einfall:
@jens-maus kommt das feature noch :-) ?
Danke für deine Arbeit !

jens-maus:
@einfall Definitiv, ja. nur dauert es eben weil es dann auch noch mit einer art Rescue-System kommen wird wie bei einer CCU2 sodass man sogar bei defektem Root-FS sein RaspberryMatic trotzdem wiederherstellen lassen kann.

robdg80:
Nut zur Info: Es gibt bereits [hier](https://github.com/j-a-n/raspberrymatic-addon-rmupdate) ein AddOn, welches ein Auto-Update von RaspberryMatic ermöglicht. Vielleicht kann man mit dem Autor zusammenarbeiten um arbeit zu sparen.

jens-maus:
Alles schon passiert (mehr oder weniger). Man beachte auch den Namen des Autors dieses Tickets hier :)

Allerdings muss man dazu auch sagen, dass technisch das Addon einen anderen Weg verfolgt als die geplante integrierte Funktionalität haben wird. Wie schon erwähnt wird es dann auch ein Rescuesystem geben über das nicht nur die Updates laufen werden sondern dieses auch erlaubt selbst bei komplett defektem Root-Filesystem RaspberryMatic wieder lauffähig zu bekommen. Das alles wird exakt der Funktionalität der CCU2 selbst nachempfunden werden.

rene-re:
Ein Auto-Update via GitHub klingt doch sehr komfortabel. Weshalb soll ich eine zip-Datei zunächst herunterladen, um sie danach wieder hoch zu laden? Aus Usability-Sicht klingt das sehr antiquiert und bei Routern wie AVM oder anderen Geräten ist solch ein Auto-Update längst normal.

Man kann ja zusätzlich das manuelle Uodate via zip-Datei anbieten.

jens-maus:
@rene-re Dann freue ich mich schon auf dein pull-request dazu ;)

rene-re:
Hallo Jens, sehr gerne sobald ich eure Code-Qualität erreiche ;) Schätze eure Arbeit sehr und möchte lediglich anmerken, dass die bisherige, tyoische Funktionsweise des Updates nicht das Maß der Dinge ist. Inwiefern es im Scope ist hier darüber hinaus zu gehen, liegt nicht in meiner Entscheidung.  Danke für die tolle Leistung bisher! 👍

PR #132, opened 2017-08-03
fixed DS1307/DS3231 initialization

jp112sdl:
DS3231/DS1307 was not available after boot.

`hwclock: can't open '/dev/misc/rtc': No such file or directory`

### DIFF
--- a/buildroot-external/overlay/base/etc/init.d/S00eQ3SystemStart
+++ b/buildroot-external/overlay/base/etc/init.d/S00eQ3SystemStart
@@ -173,11 +173,13 @@ init_rtc_device() {
   if [[ $? -eq 0 ]]; then
     # check for DS3231
     modprobe i2c:ds3231
+    echo ds3231 0x68 >/sys/class/i2c-adapter/i2c-1/new_device
     [[ -e /dev/rtc0 ]] && return 0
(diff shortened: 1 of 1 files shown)


jens-maus:
Thanks for the PR. But I wonder which RTC Clock are you exactly using? Do you have a link to the manufacturer? I ask because I remember that I just removed these lines during the last release cycle because here in my tests they actually produces additional warning AFAIR.

jp112sdl:
It's this one:
http://www.ebay.de/itm/RaspDuino-RTC-DS-3231-Clock-Module-fuer-Arduino-Raspberry-PI-blue-/263028589588?

Without the additional line, the output of my PR `i2cdetect -y 1` says `68`
After doing the ` echo ds3231 0x68 >/sys/class/i2c-adapter/i2c-1/new_device` i2cdetect comes with `UU` and `hwclock -r` shows the time.

jens-maus:
See here for the commit where I actually removed these lines due to recognizing that they are essentially not required in my setup:

https://github.com/jens-maus/RaspberryMatic/commit/45cc4e2e948049cf1e6457a60246822eecf910ce

The question now is why they seem to be required in your case? I will see if I can reproduce the case again where these lines caused additional error messages in my setup.

Issue #203, opened 2017-11-30
Restore of sbk-file asks for security key even though none was set


hicksel:
Hi,
First of all I want to thank everyone involved for this great software.
I moved from a CCU2@2.29.23 to Raspberrymatic on a PI1B@2.29.23.20171022 and now to a PI3@2.29.23.20171118. The switch from CCU2 to Raspberrymatic worked fine. But when importing my Backup from the PI1 to the PI3 I was asked for a security key even though I had never set one. Also just entering a new combination like 12345 failed with an error message telling me to look into the service reports for a detailed error information. Of course there were no error messages. After creating a diff between the config folders from the old to the new backup I noticed that only a few config files had changed since I renamed some components. But the files which could be related to the security key like ssh and shadow files were identical between both backups. The only other differences in the archive were the signature and the firmware version. Since the signature obviously had to be different I tried changing the firmware version to 2.29.23 and now I could restore the backup without a password prompt. So there seems to be the bug that the software expects a security key to be set when importing a backup from a previous Raspberrymatic version.

jens-maus:
This is the first time I heard about such a problem. However, to further investigate the issue it would be good to see the previous content of the `firmware` file before you modified it.

hicksel:
Sorry if my description of the solution was unclear. I did edit the version info in the sbk file and not in the firmware.

jens-maus:
I understood that. But please present the version info file before and after you edited it as I am not able to reproduce the problem here.

hicksel:
Even though I don´t think there is anything special about removing the raspberrymatic version substring from the firmware_version file but I have added a zip with both versions from the sbk archive.
[changes.zip](https://github.com/jens-maus/RaspberryMatic/files/1519415/changes.zip)

jens-maus:
Sorry, but I just tried to reproduce the problem without success. Are you able to reproduce the problem yourself?

hicksel:
Sorry after trying to reproduce the issue I noticed that the prompt for the security key was a quirk of Firefox in combination with small displays that led to focusing on the set a security key section if your backup filename was too long. But still when trying to enter the security Key "Aa1234" I just get the error in the screenshot.
![set_key_error](https://user-images.githubusercontent.com/10339125/33455091-4c13ab10-d61b-11e7-8e2b-1bc7f95b5603.PNG)

jens-maus:
[...]

PR #204, opened 2017-12-02
Security Key can only be set with a working BidCos-RF interface

hicksel:
Expanded the Limitations Section

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ The RaspberryMatic project is a collaborate effort to provide a lightweight, [Li
 * No HomeMatic-IP support with HM-MOD-RPI-PCB on ASUS Tinkerboard (work in progress)
 * No Bluetooth and onboard-WiFi support for ASUS Tinkerboard (work in progress)
 * A used [HM-Sys-sRP-Pl](https://www.elv.de/controller.aspx?cid=683&detail=10&detail2=731493) repeater cannot be reconfigured
+* A Security Key can only be configured if you have a working BidCos-RF interface(HM-MOD-RPI-PCB or LAN Gateway)
 
 ## :computer: Requirements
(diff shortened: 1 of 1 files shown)


jens-maus:
Thanks for the PR. However, I am not sure if adding this is really required. And if so it should IMHO be a general comment that for some functionality working RF Hardware is required (which in fact should be obvious)

litti:
By the way: did someone test that ReGa only accepts the key if needed hardware is present and setting the key lacking needed hardware does not lead to an inconsistent state?

hicksel:
I totally agree with you that there should be a general note of all functionality that is affected by not having an RF Interface. But the reason for bringing this up is that in an Homematic Wired only Setup which I currently have not setting being able to setup a security key also means no password protection for the backup as far as I understand it. In regards to @litti I have tried to set a security key without working RF hardware and have gotten the error notice in the screenshot in issue #203. But apart from that everything seems to still work fine.

jens-maus:
The corresponding line responsible for the error is the following:

https://github.com/eq-3/occu/blob/master/WebUI/www/config/cp_security.cgi#L51

Here the `xmlrpc` command returns an error if no binrpc is listening due to the fact that no RF hardware was found and thus no `rfd` is running. Of course you can add an `if()` around that, but what would happen if rfd is not running, you change the security key via WebUI and later rfd is running, then all devices won't have the correct security key embedded and thus RF communication would not work.

hicksel:
If I understand this correctly then `rfd` is responsible for setting and storing the key for the RF interfaces. One thing that would be interesting to know is if the backup encryption also depends on `rfd` or if this is done separately. Even though I don´t know about the technical details of the `rfd` daemon would it be possible to launch it without any "real" hardware attached, then set the password and stop it again? Or alternatively change the stored key without starting `rfd`.

jens-maus:
There is no backup encryption.

PR #267, opened 2018-02-25
Modify rule to prevent automount of system-related filesystems on usb…

j-a-n:
… storage.


j-a-n:
Thanks for the quick merge Jens!

PR #349, opened 2018-06-05
Update: updateDCVars.tcl

HMside:
This patch will fix the following issues:
- Double runs or logging outputs if more than one Gateway in use
- System variable is not created because another object with the same name already exists (now use ID_SYSTEM_VARIABLES)
- Too many logging entries (now only create a entry if DC equal or greater 80 %)
- Removes unused parts / Code cleaned up

This patch will add:
- Checking if Wired-Gateway in use, then create a system variable named "Wired-Status" with the states "online/offline". This will allow as with the RF Gatways to respond via central program if the Network connection of the Wired-Gateway was disconnected


jens-maus:
Danke Andreas für diese Anpassungen!

PR #405, opened 2018-09-19
The fstab should have an fsck (if needed)

pfetzing:
According to the manpage of fstab, setting the pass to 2 should have fsck check the filesystem after other filesystems.
Setting pass to zero means fsck will not check this filesystem.

I have successfully tested this on my Raspberry Pi as it previously had issues when losing power and would require me to manually check the userfs filesystem and repair errors. After setting pass to 2 it didn't require that any more.


jens-maus:
Thanks for this little nice fix. This really seems to be an oversight.

Issue #411, opened 2018-09-23
add SortByName() method to enum UserSharedObjects()


jp112sdl:
Users can add system variables to the right side of the start page.

These system variables are sorted chronologically how they were added.
Maybe it would be clearer to order them by name.

The easiest way could be by adding the `SortByName()` method to the enum `UserSharedObjects()`, that does not work at the moment.

See [startpage.htm#L225](https://github.com/eq-3/occu/blob/5c909fa587ad7dee4670f01f1014cdc0ad564c3c/WebUI/www/rega/pages/tabs/startpage.htm#L225)

jens-maus:
@jp112sdl Thanks for the suggestion. However, the object UserSharedObjects() returns is formally no Enum, it returns an `idarray` and for this ReGaHss object type no `SortByName()` implementation exists yet. While I could of course add this sort function, this would require a new ReGaHss version to be published first. Thus, this could in fact take some time to be actually implemented.

jp112sdl:
I think that is not a problem. The list is unsorted all the years... No need to hurry :)

jens-maus:
see https://github.com/jens-maus/RaspberryMatic/commit/8003adba4f652d51f6ed53aca732eee55f1d7428

jp112sdl:
@jens-maus 
Hi, kannst du sagen, ob diese Änderung in der [ReGaHss.community](https://github.com/eq-3/occu/blob/2c6aeffbbf1ea752895e845939a832d4570e5596/arm-gnueabihf/packages-eQ-3/WebUI/bin/ReGaHss.community) des OCCU Commit [#2c6aeff](https://github.com/eq-3/occu/commit/2c6aeffbbf1ea752895e845939a832d4570e5596) enthalten ist?
Im Kommentar wird nur auf #441 Bezug genommen.

PR #448, opened 2018-10-26
add webui patch file to sort user systemvariable favorites by name

jp112sdl:
...the last piece to implement suggestion in #411


jens-maus:
Great. Thanks. But I would actually prefer if we could implement sorting not on each display of the usersharedobjects list but when actually adding/removing items in the webui (user configuration). There we could then also add functions to actually manually sort the list because many users actually want to sort it manually and not just alphabetically.

jens-maus:
You can simply regenerate the idarray as a new copy of the old but in a sorted fashion. IMHO there should be up/down buttons added to perform this operation and perhaps a sort button to sort alphabetically. And I think an survey should not be necessary.

jp112sdl:
Now, the user can sort the items on demand by clicking a button.

bildschirmfoto 2018-10-26 um 22 10 33

PR #562, opened 2019-02-24
add hint that HmIPW-DRAP only supports homematicIP-Wired...

hoedlmoser:
…together with RPI-RF-MOD


jens-maus:
I took the #563 PR instead.

PR #563, opened 2019-02-24
replace listing of supported devices by table

hoedlmoser:
its up to you, Jens, if you prefere a listing or an overview table.


jens-maus:
Thanks. I will take this and adapter it accordingly.

PR #568, opened 2019-03-01
direct ntpdate to use an unprivileged port ...

hoedlmoser:
... for outgoing packets, as chrony already does. this should prevent NTP packets get filtered by some snake-oil security filtering on UDP packets with same src and dst port.


jens-maus:
Thanks for this nice patch! Didn't know that ntpdate has such an option...

PR #597, opened 2019-04-09
Protection against XSS by CSP & mask ReGaHss Servername

psytester:
1. the setenv.set-response-header part will overwrite the possibly provided ReGaHss Server name on Lighttpd controlled ports
The HTTP response header will not longer contain ```Server: ise GmbH HTTP-Server v2.0``` --> it becomes ```Server: Server```
If a system is available on open port 80 by portforwarding, it is not longer identifiable by this special server name
Currently there are 8303 open HomeMatic systems listed: https://www.shodan.io/search?query=ise+GmbH+HTTP-Server+v2.0
It's (a little bit) harder for the attacker to find a HomeMatic system, if the Server HTTP header is obfuscated.

2. The CSP setting protects against XSS attacks. It needs to allow remote host ```*.homematic.com``` to be able to load version check java script from http://ccu3-update.homematic.com/firmware/download?cmd=js_check_version&........
You may adjust the host name for RaspberryMatic update check, too.

### DIFF
--- a/buildroot-external/overlay/base/etc/lighttpd/conf.d/setenv.conf
+++ b/buildroot-external/overlay/base/etc/lighttpd/conf.d/setenv.conf
@@ -1,4 +1,10 @@
+setenv.set-response-header = (
+  "Server" => "Server"
+)
+
 setenv.add-response-header = (
+  "Content-Security-Policy" => "default-src 'self';frame-ancestors 'self';script-src 'unsafe-inline' 'unsafe-eval' 'self' *.homematic.com;style-src 'unsafe-inline' 'self';img-src 'self' data:",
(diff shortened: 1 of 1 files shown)


jens-maus:
As your PR is for RaspverryMatic please provide the correct CSP setting and please provide references to appropriate documentation explaining the theory behind your changes in detail.

psytester:
I don't know the update check URLs used in RaspverryMatic, as the maintainer you should adapt them.

Which theory do you mean? The configuration usage or the help of response obfuscation?
Since 1.4.46 lighttpd is able to use the setenv.set-environment() refer to 
https://redmine.lighttpd.net/issues/650
https://redmine.lighttpd.net/issues/2295
https://redmine.lighttpd.net/projects/1/wiki/Docs_ModSetEnv

jens-maus:
Thanks for your PR. After some checks I integrated it now and also enhanced it a bit.

PR #691, opened 2019-08-26
Mitigate another RCE possibility

psytester:
As communicated the access to root URL should be allowed for HTTP GET method only.
This will mitigate the proven RemoteCodeExecution, otherwise the next check
```$HTTP["url"] !~ "^/(config/)|(upnp/)|(webui/)|(ise/)|(api/)|(tools/)|(pda)|(pages/jpages)|(addons).*" {```
will apply, because "!~" means "expression not match" and finally the request with its payload is forwarded to ReGa process

Disclaimer:
The start page of WebUI is accessible on ```http://1.2.3.4``` and ```http://1.2.3.4/``` in browser by common HTTP GET request and will redirect to login page. All other HTTP Methods are blocked.
The check could be enhanced by checking against localhost vs. remote IP.
I did not check any negative impact to existing external tools, AddOns nor scripts.

### Types of changes

### Contributing checklist

- [ ] I have read the **CONTRIBUTING** and **LICENSE** document.
- [ ] I fully agree to distribute my changes under Apache 2.0 license.

### DIFF
--- a/buildroot-external/overlay/base/etc/lighttpd/conf.d/proxy.conf
+++ b/buildroot-external/overlay/base/etc/lighttpd/conf.d/proxy.conf
@@ -3,8 +3,15 @@
 #-----------
 $SERVER["socket"] == ":80" {
   server.document-root = server_root
+  # Deny external access to such URLs, allow localhost access within HM scripts only
   $HTTP["url"] =~ "^/.*\.(exe|oxml|hssml).*" {
     $HTTP["remoteip"] !~ "^(127\.0\.0\.1|::ffff:127\.0\.0\.1|::1)$" {
+      url.access-deny = ( "" )
+    }
+  }
+  # Mitigate another Remote Code Execution possibility, access to Root URL only allowed with HTTP GET method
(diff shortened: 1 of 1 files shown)


jens-maus:
@psytester Thanks for the PR. However, I am currently working on fixing the root of issue in ReGaHss directly as there are more RCE issues without valid sessionIDs one can execute on subpages (e.g. `/esp/system.htm`) as I just identified myself.

psytester:
:+1:
By definition, you are now a security expert. :wink:

psytester:
Upp, thank you to point to the issue...
I can confirm that some of those URLs are vulnerable too.
This PR makes no sence anymore.
But I keep open this PR until you provide the fix of the root cause.

jens-maus:
So, the issue this PR should fix is now fixed in ReGaHss R1.00.0388.0212 and newer version. See here for a rc/test version with this ReGaHss version included which should hopefully fix all known CVEs and security issues we were talking about :)

https://homematic-forum.de/forum/viewtopic.php?f=65&t=52764&p=526372#p526364

PR #692, opened 2019-08-29
Force Cache-Control for all mime types

psytester:
### Description

This should minimize the display errors in browser after software updates.
The subsequent ReGa process does not set the cache control header for all MIME types and so the CSS and JS files are cached in the browser.
This setting forces the use of this HTTP header and overwrites possible values.
After cleaning the cache once, the browser no longer caches the data and always reloads it.

Disclaimer:
Stability vs. speed?
I am aware, that this will increase the loading time, because now all files will be fetched fresh.
I have not checked any negative effects on existing external tools, add-ons or scripts.

### Types of changes

### Contributing checklist

### DIFF
--- a/buildroot-external/overlay/base/etc/lighttpd/conf.d/setenv.conf
+++ b/buildroot-external/overlay/base/etc/lighttpd/conf.d/setenv.conf
@@ -1,5 +1,6 @@
 setenv.set-response-header = (
   "Server" => ""
+  "Cache-Control" => "no-store, no-cache, max-age=120"
 )
 
 setenv.add-response-header = (


jens-maus:
Thank for your PR on browser caching improvements for the WebUI. However, please note, that we already have IMHO some good improvements regarding browser caching in RaspberryMatic. See here:

https://github.com/jens-maus/RaspberryMatic/blob/master/buildroot-external/patches/occu/0026-WebUI-Fix-CacheControl.patch

This WebUI patch introduces some additional URL parameters (`?_version_=XXX-WEBUI-VERSION-XXX`) with the actual firmware version being put where `XXX-WEBUI-VERSION-XXX` is referenced. This URL parameter is added for all javascript + css (and some cgi) file references in various html pages of the WebUI (see patch). And as modern browsers are checking the whole URL including parameters in their caching mechanisms this in fact causes the browser to NOT load a certain javascript/css/cgi file from the cache with every new version being released.

And since we introduced this optimization we didn't see any further browser caching related issues in the WebUI of RaspberryMatic. Please note, however, that this is currently a RaspberryMatic patch only and neither the CCU3 or CCU2 WebUI is using this mechanism.

In addition, I still feel that this might be a better approach than simply forcefully disabling browser caching completely for the CCU like you are proposing here. Opinions welcome, thought.

psytester:
You're right my method is the hard one.

Until now I understand the cache control in the browser, that the filename has to change for the file to be reloaded. This will work in 100%
(https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/http-caching?hl=de)

Oh, wonder, version query does the same.
(https://medium.com/@codebyamir/a-web-developers-guide-to-browser-caching-cc41f3b73e7c)
A query on a changing "?_version_=..." value does the same, but has some drawbacks if we want to cache the files but a proxy is used. Uncommon in this setup.

If RaspberryMatic runs without errors now as all related HTML pages using JS & CSS import were enhanced by the version query, the PR can be closed.

PR #788, opened 2020-02-26
suggestion for #779 and #785

jp112sdl:
Suggestion for #779 
- implement a new css-style `.tBodyCellInactive` with a light-grey color (uses `inactiveBackground` from color.map) 
- set `<td>`-Class to `.tBodyCellInactive` if program is inactive
Bildschirmfoto 2020-02-26 um 09 01 42

**known issue:**
- changing the "active" state will not apply the style immediately. you have to reload/recall the program list page

and #785
- add the name of the program to delete in popup
Bildschirmfoto 2020-02-26 um 09 02 24

_maybe we should put the name to the caption of the popup?_


jens-maus:
Thanks for the PR to address the two WebUI suggestions. However, can you please generate separate PR for each single suggestions so that we can discuss them separately. That would be great!

jp112sdl:
done with #789 and #790

PR #789, opened 2020-02-26
suggestion for #779

jp112sdl:
Suggestion for #779 
- implement a new css-style `.tBodyCellInactive` with a light-grey color (uses `inactiveBackground` from color.map) 
- set `<td>`-Class to `.tBodyCellInactive` if program is inactive
Bildschirmfoto 2020-02-26 um 09 01 42

**known issue:**
- changing the "active" state will not apply the style immediately. you have to reload/recall the program list page


jens-maus:
Thanks for this PR!. This almost looks like what #779 suggested. Regarding the known limitation (refresh required) have you tried to use something like a table jQuery refresh upon active/inactive checkbox press. See here for example:

https://datatables.net/forums/discussion/38969/reload-refresh-table-after-event

jp112sdl:
Das ist für DataTables... da hört mein rudimentäres Verständnis für Web-Programmierung gänzlich auf :/ 
Aber vielleicht hat jemand aus der Community die Möglichkeit, sich der Sache anzunehmen.

PR #801, opened 2020-03-09
webUI - add a copy to clipboard button for device addresses in the device list

Hoytron:
### Description

WebUI feature:
Add a Copy-To-Clipboard Button to the serial number in the device list.  

Problem: 
You can not copy the serialnumber from the device list, because of the immediate click event after you have selected the device serialnumber. 

Selected Use Cases:
- When you write ReGa scripts and use the serialnumber to address devices
- When you try to debug and need to search for serialnumbers in log files 

Proposed Solution: 
A small copy to clipboard button within the device list allows the user to copy a serialnumber to his/her clipboard without opening the general device settings modal/overlay

example

### Types of changes

### Alternate Designs

- no alternate designes were considered
- proposed solution is a minimal change without usage drawbacks

### Verification Process

- build the image (for raspberry pi 4) and tested it 
- test with a device list of multiple devices
- test with Firefox73, Chrome80, Edge80, IE11
- test with different display resolutions and zoom 

### Release Notes

- [WebUI] New copy-to-clipboard button for device serialnumbers 

### Contributing checklist

### Licensing conflicts
Not applicable


jens-maus:
Sorry, but I am not sure if this clipboard button is really necessary because it is easily possible to copy the serial number of a device to the clipboard by selecting the serial number in the device information dialog when clicking on the corresponding device in the device list,

Hoytron:
Unfortunatly it's not possible for me to select the serial number anywhere in the devicelist without opening any overlay or triggering a drag'n'drop action.
The serial number in the information dialog of a device is unselectable. maybe you consider to allow a selection of the serial number there.

jens-maus:
@Hoytron Which "information dialog" are you talking about? Here in my tests I can select the serial number by fist clicking on a device and then select the serial number from the popup/overlay, then copy&paste it where I want and close the overlay again. IMHO this should be sufficient.

mifricke:
I think this improvement is absolutely useful! In fact, I cannot copy the number without opening the dialog. It's not essential for life, but the change seems to be finished and doesn't hurt anyone...

An integration also motivates the submitting developer to create more WebUI improvements in the future.

jens-maus:
@mifricke I agree that the change is finished and in principle working. And of course I didn't want to demotivate @Hoytron at all. In addition, patches to the main `webui.js` file are always problematic and a require close consideration because they are prone to produce severe merge conflicts with the other patches we provide for the WebUI. So they are not as unproblematic like patches to other files.

Furthermore, all I want to achieve is to get the best implementation possible and in fact, his proposed method only helps to copy the serial number into the clipboard. What's about the Name, Description, etc. fields? In fact, if one could provide a patch that actually changes the device list so that in general all information can be easily copied to the clipboard (e.g. by providing a different possibility to actually show the device properties (perhaps by just pressing on the device name column?) because this is the main reason why a simple "select + cmd+c" doesn't work) this would be IMHO the better and more viable solution.

jens-maus:
Sorry, but after having thought about this PR for a long time now, I still feel that is has little to no real impact and that it does not justify the changes that are required. So sorry, but I will finally reject it, also because there haven't been any real significant user feedback to it. However, please don't feel demotivated by this rejection, but please submit more different PRs after having discussed them previously via adequate issue tickets here.

Hoytron:
> 
> 
> Sorry, but after having thought about this PR for a long time now, I still feel that is has little to no real impact and that it does not justify the changes that are required. So sorry, but I will finally reject it, also because there haven't been any real significant user feedback to it. However, please don't feel demotivated by this rejection, but please submit more different PRs after having discussed them previously via adequate issue tickets here.

@jens-maus  Sorry for my late reply. No worries for rejecting this PR. I was very busy since 05/2020. 
I understand your decision for the project. A complete intuitive UI refresh woud be greate. Unfortunately a do not have the required skills for this task. 

I also would love a more general solution as you suggest for all the columns.

PR #813, opened 2020-04-19
use /bin/bash to make sure bashism's like "[[" don't fail

Yannik:
fixes
```
/usr/local/bin/install_addon: 9: /usr/local/bin/install_addon: [[: not found
/usr/local/bin/install_addon: 19: /usr/local/bin/install_addon: [[: not found
/usr/local/bin/install_addon: 25: /usr/local/bin/install_addon: [[: not found
/usr/local/bin/install_addon: 31: /usr/local/bin/install_addon: [[: not found
```
errors

### Types of changes

### Contributing checklist

- [ ] I have read the **CONTRIBUTING** and **LICENSE** document.
- [ ] I fully agree to distribute my changes under Apache 2.0 license.

### DIFF
--- a/buildroot-external/overlay/base/bin/install_addon
+++ b/buildroot-external/overlay/base/bin/install_addon
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 # simple install_addon script that will install CCU addons
 # and return either codes >= 10 to signal errors or lower
--- a/buildroot-external/overlay/base/bin/install_addon
+++ b/buildroot-external/overlay/base/bin/install_addon
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 #
 # simple install_addon script that will install CCU addons
 # and return either codes >= 10 to signal errors or lower


jens-maus:
Sorry, but your propose change doesn't see to be right. Please note:

1. your description is refering `/usr/local/bin/install_addon` but your change tries to change `/usr/bin/install_addon`
2. There is no /bin/bash in RaspberryMatic. See:
```
root@homematic-raspi:~# ls /bin/bash
ls: /bin/bash: No such file or directory
```

So please explain why you think your proposed change is required or correct.

Yannik:
@jens-maus Dash is the default non-interactive shell in debian since quite some time: https://wiki.ubuntu.com/DashAsBinSh#A.5B.5B

It does not supprt bashisms like `[[`. 

I've amended my commit to make sure it works even if `bash` has any other path than `/bin/bash`.

jens-maus:
@Yannik Sorry, but I think you are mixing something up here. RaspberryMatic isn't based on debian but uses Buildroot and uses a busybox environment without any bash shell interpreter installed. Thus, neither your `/bin/bash` nor your env-based change is correct here. My impression is that you don't even use RaspberryMatic but something like piVCCU or Debmatic, which is fundamentially different.

Yannik:
@jens-maus Ok, sorry. I was under the impression raspberrymatic used debian too. I wanted to use your script in my own [setup](https://github.com/Yannik/ansible-role-occu) and make it more portable in the meanwhile, but I guess I'll just fork it. Thank you!

jens-maus:
@Yannik nope. neither RaspberryMatic nor the CCU3 operating system are based on debian/raspbian and thus are not coming with bash. Thus, if you try to install a OCCU environment in a full-fledged debian/raspbian system you need to adapt things on your own and make sure to map pathes like /www/config which are using in several different places to the expected pathes similar to RaspberryMatic/CCU3. This is, e.g., exactly what piVCCU or debmatic is doing, btw!

PR #1056, opened 2021-01-03
[WIP] Initial support for containers

angelnu:
### Description

Adds support to build Open Container Initiative ( as Docker) containers.

### Related Issue

https://github.com/jens-maus/RaspberryMatic/issues/786

### Types of changes

- [X] New feature (non-breaking change which adds functionality)

### Alternate Designs

- building just the tarball and keeping the docker wrapper in current angelnu/ccu git repo.
- extract files from ova images using qemu tools

Both are circumvention since they would patch RaspberryMatic after it has been built which makes very difficult not running in broken builds.

### Possible Drawbacks

Kernel headers must be kept as old as possible to support running on host with older kernels. For example the Sinology NAS is still using kernel 4.4 which is the oldest currently supported by buildroot.

The old headers are needed in the OCI images AND the multilib. Currently I did not create new definitions for multilib since I do not see what problem would be using the older headers there but it could be done if anyone sees an issue.

### Verification Process

Started the container on my amd64 workstation which uses kernel 4.19. All init error problems are gone and the UI comes up sucesfully. I still need to test with actual RF devices and build&test on arm.

Beyond that a CI will need to be contributed that uploads to a container repository. I propose to use https://quay.io/repository/angelnu/raspberrymatic (since docker hub new policy is not great)

### Release Notes

- Container images available for amd64, arm and arm64

### Contributing checklist

- [X] My code follows the code style of this project.


angelnu:
# Build command
`make PRODUCT=raspmatic_oci_amd64 build -j 8`

# Run command
`docker run -ti --rm --name test -v ccu_data:/usr/local -p 8080:80 -v /dev:/dev --privileged raspberrymatic:amd64-latest`

Notes:
- installed `pivccu-modules-dkms` in host.
- dev is shared to allow dynamically adding the virtual raw devices
- privileged is needed to reset the raw_uart device

# Boot output
```
Identifying onboard hardware: oci, OK
Initializing ZRAM Swap: SKIP
Initializing RTC Clock: onboard, OK
Initializing urandom: SKIP
Checking for Factory Reset: not required
Checking for Backup Restore: not required
Initializing System: OK
Starting logging: OK
Identifying Homematic RF-Hardware: .....HmRF: HM-MOD-RPI-PCB/HB-RF-USB@usb-0000:00:14.0-1, HmIP: HM-MOD-RPI-PCB/HB-RF-USB@usb-0000:00:14.0-1, OK
Updating Homematic RF-Hardware: HM-MOD-RPI-PCB: 2.8.6, OK
Starting bluetooth: disabled
Starting network: SKIP
Starting chrony: SKIP
Preparing start of hs485d: no Hm-Wired hardware found
Starting xinetd: OK
Starting eq3configd: OK
Starting lighttpd: creating new SSL cert... OK
Starting ser2net: no configuration file
Starting ssdpd: OK
Starting NUT services: disabled
Initializing Third-Party Addons: OK
Starting LGWFirmwareUpdate: ...OK
Setting LAN Gateway keys: OK
Starting hs485d: no Hm-Wired hardware found
Starting multimacd: .OK
Starting rfd: .OK
Starting HMIPServer: ........OK
Starting ReGaHss: .OK
Starting Third-Party Addons: OK
Starting crond: OK
Setup onboard LEDs: booted, OK

```

# Known bugs:
None so far - web UI comes up and I can learn new devices

jens-maus:
Thanks and welcome back again ;-) This already looks quite promising. Nevertheless, as this is a major feature please give me some time to test and integrate it thoroughly. So thanks that you added [WIP] because I think this might take some time to integrate and test it for all potential use-cases (HB-RF-USB as well as GPIO bus uses, etc. etc.) as well as testing potential issues with third-party ccu addons, etc.

angelnu:
on the container repository - I thought github was charging but it does not to open repositories - so same policy as quay.io with open source :-) I have not tried github actions yet but seems quite standard CI so I should get it figured out to add to the PR (at least you want to do that part yourself)

jens-maus:
> I have not tried github actions yet but seems quite standard CI so I should get it figured out to add to the PR (at least you want to do that part yourself)

Feel free to generate a template workflow action yml if you like. But I guess for the final github packages setup I would need to take over anyway to get it smoothly integrated into the build+development workflow here in the main repository so that the final distributed packages will be part of the jens-maus/RaspberryMatic github project. See, btw, here for some example yml files to deploy docker images to the new github container registry (ghcr.io): https://github.com/github/super-linter/tree/master/.github/workflows (I will also try to train myself the knowledge of how to publish the docker images to the new github container registry which seems to me a proper place for the potential future raspberrymatic docker images.

buildroot-external/overlay/base/etc/init.d/S49chrony
```diff
@@ -1,5 +1,11 @@
 #!/bin/sh
 
+# source all data from /var/hm_mode
+[[ -r /var/hm_mode ]] && . /var/hm_mode
+
+# oci platform does not require own urandom
```
angelnu, line 6: comment does not match

buildroot-external/overlay/base/etc/init.d/S40network
```diff
@@ -6,18 +6,17 @@
 # Debian ifupdown needs the /run/network lock directory
 mkdir -p /run/network
 
+# source all data from /var/hm_mode
+[[ -r /var/hm_mode ]] && . /var/hm_mode
+
+# oci platform does not require own rngd
```
angelnu, line 12: comment does not match

buildroot-external/overlay/base/etc/init.d/S04CheckFactoryReset
```diff
@@ -51,7 +51,9 @@ check_factory_reset() {
     # mount /usr/local again
     mount /usr/local
   else
+    # clean whole /usr/local on oci platform
     rm -rf /usr/local
```
angelnu, line 55: actually this fix was bad - it should delete the content of the volume, not the volume itself - will fix

buildroot-external/overlay/base/etc/init.d/S49chrony
jens-maus: thx, will fix

buildroot-external/overlay/base/etc/init.d/S40network
jens-maus: will fix

buildroot-external/overlay/base/etc/init.d/S04CheckFactoryReset
jens-maus: sure, but `rm -rf /usr/local/*` will omit dotfiles, so if you fix it yourself, keep this in mind ;)

jens-maus:
> I understand it will take some time - no rush - I will let it run in my K8S cluster to see how it goes but it boots/feels very fast. Ideally enabling daily builds first before making a public release so any breaking change can be caught.

This is also my plan, indeed. So first getting this thing stabilized to some extent and the publish first development docker images via github container registry so interested users can test it thoroughly.

> On tests - being docker containers it should be possible to add some automated testcases. I tried to start without any device but Rega did not boot. Maybe with the dummy raw_device? It is an impressive project with all the features and supported HW so manual testing would be a challenge....

It is challanging, yes. And indeed, after we have that thing out we can reuse these docker images for generating proper testcases. You probably have noticed, but I already have a occu testing repository which @hobbyquaker kindly developed and I took over (see https://github.com/jens-maus/occu-test). I already use that to test my changes on ReGaHss itself.

angelnu:
I created a test oci-ci branch in my fork to test there the docker build before merging into the main PR

angelnu (commented):
mounts would require the container to be privileged which is not a good practice.

I use privileged in my command until we have a fix for:
```
Identifying Homematic RF-Hardware: ../etc/init.d/S11InitRFHardware: line 179: can't create /sys/class/raw-uart/raw-uart/reset_radio_module: Read-only file system
```

jens-maus:
> mounts would require the container to be privileged which is not a good practice.

Well, then we need a different fix for the mixed /tmp, /var/run vs. /run as well as /var/log layout which is substantially different compared to a real raspberrymatic. But for the time being using privileged seems to be an appropriate workaround.

jens-maus:
> # Run command
> `docker run -ti --rm --name test -v ccu_data:/usr/local -p 8080:80 -v /dev:/dev --privileged --hostname homematic-raspi raspberrymatic:amd64-latest`

BTW: I think we also need a different approach in starting the docker, because giving it full access to the whole `/dev` tree e.g. prevents syslogd within raspberrymatic from working/starting since it want to access `/dev/log` on the host then. Thus, in the final version we have to only expose the dev nodes (e.g. `/dev/eq3loop`, etc.) that are required using the `--device=` option, I guess.

angelnu:
> 
> 
> > # Run command
> > `docker run -ti --rm --name test -v ccu_data:/usr/local -p 8080:80 -v /dev:/dev --privileged --hostname homematic-raspi raspberrymatic:amd64-latest`
> 
> BTW: I think we also need a different approach in starting the docker, because giving it full access to the whole `/dev` tree e.g. prevents syslogd within raspberrymatic from working/starting since it want to access `/dev/log` on the host then. Thus, in the final version we have to only expose the dev nodes (e.g. `/dev/eq3loop`, etc.) that are required using the `--device=` option, I guess.

The problem is that --device requires the device to exists when the container is started which is not the case for those devices added by multimac

There the option of mount the host dev under /dev_host and then `ln -s /dev_host/<dev> /dev/<dev>`. IMO is still a circumvention (ideally the udev inside the container would create the device files) but it is what I used in my original project.

jens-maus:
> > BTW: I think we also need a different approach in starting the docker, because giving it full access to the whole `/dev` tree e.g. prevents syslogd within raspberrymatic from working/starting since it want to access `/dev/log` on the host then. Thus, in the final version we have to only expose the dev nodes (e.g. `/dev/eq3loop`, etc.) that are required using the `--device=` option, I guess.
> 
> The problem is that --device requires the device to exists when the container is started which is not the case for those devices added by multimac
> 
> There the option of mount the host dev under /dev_host and then `ln -s /dev_host/<dev> /dev/<dev>`. IMO is still a circumvention (ideally the udev inside the container would create the device files) but it is what I used in my original project.

I perfectly understand that, but to me it seems rather awkward that the whole /dev is mapped into the container rather than limiting it to the dev devices required. I agree, for the devices created by multimacd this might be an issue and we could indeed introduce a /dev_host and then link to them accordingly from /dev. So if you can show me how you do it in your original project I could see if I could adapt that for RaspberryMatic.

> **EDIT**
> @jens-maus - is there a reason for using always() in the CI. This is preventing me to cancel running workflows which is bad for me to test the CI. I will disable at least temp for the _oci_ and _oci_test_ branch.

Yes, the reason is that otherwise the nightly build scheduled build won't work correctly. But for the WIP period of this PR I am perfectly fine with this and I have anyway planned to see if I can change the way the nightly builds are currently created. But I will wait for this PR to stablize and being integrated into master, hopefully soon :)

angelnu:
> 
> 
> > > BTW: I think we also need a different approach in starting the docker, because giving it full access to the whole `/dev` tree e.g. prevents syslogd within raspberrymatic from working/starting since it want to access `/dev/log` on the host then. Thus, in the final version we have to only expose the dev nodes (e.g. `/dev/eq3loop`, etc.) that are required using the `--device=` option, I guess.
> > 
> > 
> > The problem is that --device requires the device to exists when the container is started which is not the case for those devices added by multimac
> > There the option of mount the host dev under /dev_host and then `ln -s /dev_host/<dev> /dev/<dev>`. IMO is still a circumvention (ideally the udev inside the container would create the device files) but it is what I used in my original project.
> 
> I perfectly understand that, but to me it seems rather awkward that the whole /dev is mapped into the container rather than limiting it to the dev devices required. I agree, for the devices created by multimacd this might be an issue and we could indeed introduce a /dev_host and then link to them accordingly from /dev. So if you can show me how you do it in your original project I could see if I could adapt that for RaspberryMatic.

It was nothing too special: https://github.com/angelnu/docker-ccu/search?q=dev_org I did it in the entryfile.sh to avoid patching the original init.d script from the CCU2. For raspbian I would modify the `S11InitRFHardware` and `S60multimacd` to point to /dev/org.

Btw: another reason for not using --device is that is not available on Kubernetes - there you actually can (need) to write a plugin to pass/virtualize devices. So /dev/org is the most compatible option.

jens-maus:
> It was nothing too special: https://github.com/angelnu/docker-ccu/search?q=dev_org I did it in the entryfile.sh to avoid patching the original init.d script from the CCU2. For raspbian I would modify the `S11InitRFHardware` and `S60multimacd` to point to /dev/org.
> 
> Btw: another reason for not using --device is that is not available on Kubernetes - there you actually can (need) to write a plugin to pass/virtualize devices. So /dev/org is the most compatible option.

Ok, got it. Thanks for the hint. While `dev_org` reads IMHO quite awkward (and I would prefer `/dev_host` instead) I would also perfer to not search&change all places where currently `/dev` is used, but simply link (if possible) all devices we require from `/dev_host` to `/dev` to have them available in `/dev` directly. 

And indeed, we would also have to think about how users would/can start the docker container using a generalized script/startup routine similar to your `deploy.sh`. This script would the make sure to not only install all dependencies (pivccu dkms modules), but also run the docker command with the necessary cmd-line options. Kubernetes support, I would, however, keep on low priority since only experts are really using Kubernetes ATM. So we should target ordinary docker users for the moment which want to run a dockerized raspberrymatic on their amd64/arm/arm64 ubuntu/raspberrypios ATM or wanting to run it on their NAS or third-party system with docker in mind.

> > Yes, the reason is that otherwise the nightly build scheduled build won't work correctly.
> 
> What I noticed is that the nighties upload to hardcoded path so they would break in forks (so we will not see what breaks without it) and that they do not upload to the artifacts so we cannot build the containers. Is there any reason for not uploading to artifacts in all cases?

Yes, because I use the artifacts feature currently only for storing the intermediate builds so I can download them manually in case I want to inspect them. And thus, for the nightly builds I don't need them since all builds are stored in the releases section of the RaspberryMatic release pages.
 
> Btw2: I enabled the workflows in my fork to test there the oci-ci branch: https://github.com/angelnu/RaspberryMatic/actions

I saw your worklows already. Let's wait until they are done. But my suspicion is, that's its time to split this huge ci.yml into smaller pieces and e.g. do all the dockker building and deploy in a separate workflow and without having to download the intermediate artifacts like you implemented it currently.

IMHO we should now concentrate in getting the rest of the docker environment finished/done as much as posisble as we could tune/adapt the docker CI workflows at a later time quite easily (I can do that later on). So please merge your intermediate oci_ci branch into this active oci branch ASAP so that we don't divert any further and concentrate to work on a single branch here.

angelnu:
> Ok, got it. Thanks for the hint. While `dev_org` reads IMHO quite awkward (and I would prefer `/dev_host` instead) I would also perfer to not search&change all places where currently `/dev` is used, but simply link (if possible) all devices we require from `/dev_host` to `/dev` to have them available in `/dev` directly.

For adding the symlinks to /dev_host we could either modify `S01InitHost? or create a new one in the oci overlay. Any preferences?
 
> And indeed, we would also have to think about how users would/can start the docker container using a generalized script/startup routine similar to your `deploy.sh`. This script would the make sure to not only install all dependencies (pivccu dkms modules), but also run the docker command with the necessary cmd-line options. Kubernetes support, I would, however, keep on low priority since only experts are really using Kubernetes ATM. So we should target ordinary docker users for the moment which want to run a dockerized raspberrymatic on their amd64/arm/arm64 ubuntu/raspberrypios ATM or wanting to run it on their NAS or third-party system with docker in mind.

I can add a deploy.sh script that would get uploaded to the release for the basic docker setup on a Debian based system. It would, as you suggest, install the pivccu dkms module and start the container as daemon and with data volume.

Other more advanced setups or that require multiple (visual) steps such as NAS or K8S could go into dedicated readme pages.


> Yes, because I use the artifacts feature currently only for storing the intermediate builds so I can download them manually in case I want to inspect them. And thus, for the nightly builds I don't need them since all builds are stored in the releases section of the RaspberryMatic release pages.

Ok, I could add rules to download from there for nighties or to also upload the OCIs tars that are inputs for the following job also to artifacts in nighties.

> I saw your worklows already. Let's wait until they are done. But my suspicion is, that's its time to split this huge ci.yml into smaller pieces and e.g. do all the dockker building and deploy in a separate workflow and without having to download the intermediate artifacts like you implemented it currently.

The artifacts upload/download are needed in order to benefit from the parallel builds for buildroot while still building a mutiarch container that supports all required architectures.
 
> IMHO we should now concentrate in getting the rest of the docker environment finished/done as much as posisble as we could tune/adapt the docker CI workflows at a later time quite easily (I can do that later on). So please merge your intermediate oci_ci branch into this active oci branch ASAP so that we don't divert any further and concentrate to work on a single branch here.

Yes, as soon as it works once end-end, hopefully today, I will merge back. I am waiting for some of the long builds to complete so the new ones can run. This is when I noticed that I could not just abort the old ones...

jens-maus:
> For adding the symlinks to /dev_host we could either modify `S01InitHost? or create a new one in the oci overlay. Any preferences?

I would probably add them to `S06InitSystem` for a start, because this is already used for some device related setup stuff. So if you add the linking of all common devices from `/dev` to `/dev_host` this would be great. Thanks!

> I can add a deploy.sh script that would get uploaded to the release for the basic docker setup on a Debian based system. It would, as you suggest, install the pivccu dkms module and start the container as daemon and with data volume.

If you can add something like a deploy.sh script for a starting point to develop a somewhat starting script to setup all dependencies on the host and start it accordingly, this would be great. Thanks. Later on we could then even think about generating a debian-like package around that which debian-based users can then install more easy and which will install everything automatically and come with a start script that even starts the raspberrymatic transparently.

> Other more advanced setups or that require multiple (visual) steps such as NAS or K8S could go into dedicated readme pages.

I agree. For that we can then use the documentation wiki to add step-by-step screenshots like for the ova setup steps.

> > Yes, because I use the artifacts feature currently only for storing the intermediate builds so I can download them manually in case I want to inspect them. And thus, for the nightly builds I don't need them since all builds are stored in the releases section of the RaspberryMatic release pages.
> 
> Ok, I could add rules to download from there for nighties or to also upload the OCIs tars that are inputs for the following job also to artifacts in nighties.

Yes, please do so! This is hightly appreciated.
 
> The artifacts upload/download are needed in order to benefit from the parallel builds for buildroot while still building a mutiarch container that supports all required architectures.

Ah ok, now I get it. then of course it is perfectly valid to use the artifact download space for running the multiarch steps afterwards

> > IMHO we should now concentrate in getting the rest of the docker environment finished/done as much as posisble as we could tune/adapt the docker CI workflows at a later time quite easily (I can do that later on). So please merge your intermediate oci_ci branch into this active oci branch ASAP so that we don't divert any further and concentrate to work on a single branch here.
> 
> Yes, as soon as it works once end-end, hopefully today, I will merge back. I am waiting for some of the long builds to complete so the new ones can run. This is when I noticed that I could not just abort the old ones...

Ok, perfect. Would would also be good would be to also get some kind of nightly developer builds for the docker/oci platform and upload these images to the public docker registry as well, probably in a different namespace to distinguish them from the stable/release builds. And if possible, probably we should also put the image files into the `snapshots` release tag like the other nightly build platforms so that interested docker developers can more easily download and use them.

angelnu:
> I would probably add them to `S06InitSystem` for a start, because this is already used for some device related setup stuff. So if you add the linking of all common devices from `/dev` to `/dev_host` this would be great. Thanks!

Ok, adding there
 
> > I can add a deploy.sh script that would get uploaded to the release for the basic docker setup on a Debian based system. It would, as you suggest, install the pivccu dkms module and start the container as daemon and with data volume.
> 
> If you can add something like a deploy.sh script for a starting point to develop a somewhat starting script to setup all dependencies on the host and start it accordingly, this would be great. Thanks. Later on we could then even think about generating a debian-like package around that which debian-based users can then install more easy and which will install everything automatically and come with a start script that even starts the raspberrymatic transparently.

Ok, adding first draft of script and uploading to artifacts

> > Ok, I could add rules to download from there for nighties or to also upload the OCIs tars that are inputs for the following job also to artifacts in nighties.
> 
> Yes, please do so! This is hightly appreciated.

Will add
 
> Ok, perfect. Would would also be good would be to also get some kind of nightly developer builds for the docker/oci platform and upload these images to the public docker registry as well, probably in a different namespace to distinguish them from the stable/release builds. And if possible, probably we should also put the image files into the `snapshots` release tag like the other nightly build platforms so that interested docker developers can more easily download and use them.

In progress :-)

angelnu:
@jens-maus - you will have to create a Personal Access Token for the Container Repository called `CR_PAT`. See https://docs.github.com/en/free-pro-team@latest/packages/guides/migrating-to-github-container-registry-for-docker-images#authenticating-with-the-container-registry 

I have configured to build&push the containers for schedule and push events but not for PRs (since the secrets do not get passed to PRs from forks). You can see how they look at https://github.com/users/angelnu/packages/container/package/raspberrymatic

jens-maus:
This looks very good @angelnu. Thanks again a lot for all your work to get the RaspberryMatic docker story finally going 👍 

Now with the new direct mounting of `/dev_host`.  Can you please adapt the documentation/cmd-line in your above comment (https://github.com/jens-maus/RaspberryMatic/pull/1056#issuecomment-753647492) until you have the example `deploy.sh` ready.

So what is then actually missing what might prevent this PR from being merged to master? AFAICS the principle docker functionality seems to work nicely so I could merge this to master to get it tested by some users and then we/you can send in other PRs for adapting some specific things if required. What do you think?

angelnu:
Thanks @jens-maus for all your work in RaspberryMatic and being open for working with me in this big feature!

I think the only two missing parts are the deploy.sh and a Wiki page to document the advanced setups. I will update the comment above for the time being so IMO this PR could be merged.

Then we need testers :-) Would you announce this as alpha platform somewhere?

I would also like to announce in angelnu/ccu that I plan to discontinue support for the containers built on the official CCU and point people to RaspberryMatic as the successor. If anyone would have interest on the official one I would remove all my previous webui patches which tend to break when eq3 release new versions. But IMO I do not see too much reason to use the original one when RasberryMatic is a superior superset.

jens-maus:
> I think the only two missing parts are the deploy.sh and a Wiki page to document the advanced setups. I will update the comment above for the time being so IMO this PR could be merged.

Ok, perfect. Then I just wait for the CI checks to finish and then I will squash+merge your PR on the master and will apply some streamlining, if required. Thanks again, and please continue to submit enhancements and potentially required updates to the docker stuff of RaspberryMatic using future PRs. I will definitely review them quickly and merge them with priority.

> Then we need testers :-) Would you announce this as alpha platform somewhere?

Yes, I will announce this in the respective issue ticket as well as open an appropriate thread in the public homematic-forum.de. Do you have an account there as well so I can put reference to you and give you appropriate credits? In addition, I will of course suggest you for receiving the collected bounty (even thought is is only $20 atm) linked in the issue ticket (see https://www.bountysource.com/issues/88798894-raspberrymatic-docker-support-dockerhub-repository)

> I would also like to announce in angelnu/ccu that I plan to discontinue support for the containers built on the official CCU and point people to RaspberryMatic as the successor. If anyone would have interest on the official one I would remove all my previous webui patches which tend to break when eq3 release new versions. But IMO I do not see too much reason to use the original one when RasberryMatic is a superior superset.

Thanks for the support! Indeed, I also think that with this new raspberrymatizied docker :) implementation you could perfectly end your own endeavour without regretting it. But be assured that I will be happy to keep you as the maintainer of the docker environment as much as you like. So any future ticket closely related to the docker stuff will be certainly routed to you, if you don't have a problem with that.

jens-maus:
> I also want to test that the upgrade within the container works BUT usually containers are updated from the host so Ideally the update from within WebUI should be disabled and just indicate that the contaner must be updated from the host (running deploy.sh again)

Thanks for the reminder. I will review that as soon as everything is merged on the master branch. But I cannot say yet, if I really will drop the webui based update alltogether. While for a usual docker user redeploying the docker would indeed be the right approach, for some users it might also be desirable to performa a webui-based docker update (or in case of CI testing). So lets see. But I will certainly keep that in mind, yes.

> It is also unusual running a terminal with password in a container which cannot be accessed in daemon mode. It does not hurt too much but it adds unneeded stuff to the log (and consume some small amount of mem?). Any reason to keep it?

It was a mutual thing to enable it when you disabled the getty stuff :) But I agree, running the raspmatic-docker as a daemon doesn't make sense to keep the `/bin/login` in `/etc/inittab`. But please hive me some time to think about it once more and see if there might be some reasons to keep it. But at least I would keep it as a comment in `/etc/inittab` so that one can know easily how to re-enable it.

angelnu:
My user in the homematic forum is _vegetto_ I am not very active there since I usually focus my activity directly in github. But I would register for updates in the docker threat.

I am happy to keep supporting the docker variant so sure, assign me any issue open in GitHub. I will try to add some testcases where after a sucessful build I would try to start the container and check that the WebUI comes online. But this would go into its new PR :-)

On the updates: I modified the script called during the updates so that hopefully it works - I have not tested it yet. For plain docker users should work and advanced ones should know already so IMO we do not need to disable the WebUI and just document in the corresponding Wiki page.

And sure, you decide on the login - when you re-add it I was not sure if something in Raspberrymatic depended on it so I did not try it for that reason. I am not blocked so I did not raise up at that point.

nicx:
Great news, thanks for that!

PR #1089, opened 2021-01-22
cache containers

angelnu:
### Description

Use cache to avoid timeouts when downloading a container shortly after uploading. It also allows running the CI without providing a github secret to upload a package.

### Related Issue

NA

### Types of changes

### Contributing checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -157,6 +157,13 @@ jobs:
     steps:
     - uses: actions/checkout@v2
 
+    # In this step, this action saves a list of existing images,
+    # the cache is created without them in the post run.
+    # It also restores the cache if it exists.
+    - uses: satackey/action-docker-layer-caching@v0.0.11
+      # Ignore the failure of a step and avoid terminating the job.
+      continue-on-error: true
+    
     - name: Setup Environment
       run: |
         echo "BUILD_DATETIME=${{ needs.build.outputs.build_datetime }}" >> $GITHUB_ENV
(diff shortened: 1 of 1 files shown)


jens-maus:
Are you sure that this additional caching won't interfere with the actions/cache uses in the build step of the workflow? Because there we use actions/cache already for storing/restoring the buildroot ccache to speed up build tim? Because the maximum size of cache GitHub allows is 5GB per repository AFAIK.

angelnu:
I do not believe they conflict - they should upload to different folders in the cache.

Since the OCI images are ca 150 MBs each then we are talking above 450 MBs max . Do you know if there is a way to check how much of the cache is being used?

I considered just doing a docker save as you proposed but it does not work with multi-arch -> it just captures the current arch. This might not be critical since I had to disable the test on qemu for ARM but it would be later if we want to test on all platforms.

jens-maus:
Ok, I accepted this PR. Let's see how good it will perform and work out. The last nights' snapshot couldn't finish in time btw, because the ccache of the rpi4 build was already overwritten/wiped because of the maximum space allowed for actions/cache use. Perhaps I really have to switch to using own runners instead of using the GitHub runners...

angelnu:
the upload cache should take care of deleting older ones or the CI might be set as ignoring errors.

If problems arise I can redo the test to use a manual save&upload to artifactory.  It is more verbose and would miss the multi-arch but doable.

Own runners will be interesting to add some Homematic HW to them so real tests could be performed so in the long term it would be a good thing to consider. What model would you have in mind: people donating on runners or you buying them (with  eventual donations)?

PR #1132, opened 2021-02-16
Support load balancers and overriding image tag

angelnu:
### Types of changes

### Contributing checklist

### DIFF
--- a/helm/raspberrymatic/templates/deployment.yaml
+++ b/helm/raspberrymatic/templates/deployment.yaml
@@ -28,7 +28,7 @@ spec:
         - name: {{ .Chart.Name }}
           securityContext:
             {{- toYaml .Values.securityContext | nindent 12 }}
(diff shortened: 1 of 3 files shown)


jens-maus:
Can you please explain at least a bit what this is all about :) I am really a newbie regarding helm charts+kubernetes but would like to understand what these templates are about, who are using them and what your changes are support to achieve.

angelnu:
There are two changes in this:
- it allows to modify which version to use of the docker contain - previously it was hardcoded to the version in the chart and now that is just a default -> this is a minor feature that helps when testing different versions and also when monitoring if new versions are available (it is a full world but if you want to see the context -> https://toolkit.fluxcd.io/)
- fixed the bug in service.yaml where I had put TCP and UDP ports together -> this works usually ok until you try to put behind a load balancer (K8S does not allow to combine TCP and UDP). This broke me since Load Balancers are also used to access non-https ports directly and not go through an ingress reverse proxy.

PR #1136, opened 2021-02-17
Increase WebUi performance in regards to Translation

theimo1221:
### Description

Especially if you have plenty of devices, switching to the `Settings -> Devices` page sometimes takes 5+ Seconds.
Reviewing the current code execution with the developer tools in the browser, I could spot some weaknesses in regards to performance.
This Patch addresses those time consuming computations, by

- ...reducing computations within loops
- ...caching values where possible
- ...using vanilla js instead of JQuery at performance critical loops

### Related Issue

As performance optimization isn't necessarily an issue I didn't open an issue.

### Types of changes

It is more or less a refactoring to increase the performance.
If you would classify it as a bugfix it still would be non-breaking change.

### Alternate Designs

I first thought about adding an 1-Dimensional Dict within the `translate.js` file, to only cache for current selected language, but that way we loose all cached data, while switching back/forth through langs.

### Possible Drawbacks

- Using vanilla js instead of JQuery might decrease readability for other developer
- The additional Translation Cache will increase pages memory consumption client side slightly

### Verification Process

Using browsers override functionality, I switched between the different pages of WebUi. Content gets translated as before, but way faster.

### Release Notes

- Increased overall WebUi performance (especially Settings->Devices) due to changes in translation.

### Contributing checklist


theimo1221:
Apparently there are some places where `<br/>` is used within Translations, which doesn´t yet work with my solution. I'll take a closer look.

theimo1221:
> Apparently there are some places where `<br/>` is used within Translations, which doesn´t yet work with my solution. I'll take a closer look.

With 9fb4561 the translation logic now checks for potentially included HTML code, and using `.innerHTML = ` instead of `.textContent` for those.

Testing with the build provided by CI/CD Chain I could verify huge performance increase already.

jens-maus:
First of all: Thanks very much for this nice first-time contribution. I am currently in the process of reviewing it and good that you already found a regression of your suggested changed. And perhaps @jp112sdl and @Baxxy13 could also help in reviewing and testing it thoroughly since they have been involved in previous WebUI patch integrations and could help here.

I am really looking forward to integrate your PR as soon you found it stable enough for integration since I also believe that there is a lot of potential for speed improvements as well as for slight feature enhancements in regard to the WebUI. And seeing that an experienced Web-developer like you hopped on the topic of getting some things improved in the WebUI makes me quite happy and I would really like to encourage you to continue to review potential WebUI improvements since @jp112sdl and myself are not really web developers and would like to invest more time on other areas in case someone like you jumps on the task to improve the webui. So please continue with your submissions and also please review the WebUI related project section where I started to collect potential tickets for improvements and fixes regarding the WebUI (see https://github.com/jens-maus/RaspberryMatic/projects/3). Perhaps you will identify more areas where you could easily contribute which will be highly appreciated!

jens-maus:
> > Apparently there are some places where `<br/>` is used within Translations, which doesn´t yet work with my solution. I'll take a closer look.
> 
> With [9fb4561](https://github.com/jens-maus/RaspberryMatic/commit/9fb4561777fe5d54e8b49f15aa4ab0b86107a5cc) the translation logic now checks for potentially included HTML code, and using `.innerHTML = ` instead of `.textContent` for those.

Thanks. As said, I will try to review this PR ASAP and report back if everything works as expected and potentially @jp112sdl and @Baxxy13 could also test it and report back.

> Testing with the build provided by CI/CD Chain I could verify huge performance increase already.

I am sorry, but I had to disable the automatic CI build job for foreign PRs because we are using own self-hosted runners instead of using the GitHub runners (because they are limited in time) which generate the builds and thus running for every foreign PR is way to much load for them. However, as soon as I have integrated your PR the next nightly snapshot will include your patch and then you can use them instead.

jens-maus:
Ok, after some short tests I integrated your PR. Again, thanks a lot for this great enhancement. As we have enough time testers like @Baxxy13 could indeed test your change and report back if there are still issues!

And again, please continue submitting such great enhancements/fixes for the WebUI since we definitely lack a bunch of web developers taking care of enhancing/fixing the WebUI in general!

BTW: The only thing that you should please consider in future when sending in PRs is, that you please don't send them in based on the master branch but please create a dedicated feature branch on your fork and add the changes to the feature/bugfix branch and then send in based on that dedicated branch. This will then allow that I submit last hand changes to it before merging in into the main RaspberryMatic repository.

PR #1139, opened 2021-02-18
0112-WebUI-Fix-Faulty-Html-in-Temperatur-Lowering

theimo1221:
### Description

This is fixing an issues within the HTML Template for the "TEMPERATURE_LOWERING" as it currently results in faulty html.
Additionally I readded the `ecoOld` field with proper html, as it causes the NaN issue described below.

### Related Issue

This resolves the described bug described here https://github.com/jens-maus/RaspberryMatic/issues/1058#issuecomment-766314386 from Issue #1058.

### Types of changes

### Alternate Designs

Additionally I looked into improving `ProofAndSetValue` which I might include in a separate Pull Request.

### Possible Drawbacks

It is unclear, if there are any drawbacks due to now correct html Template. So the logic regarding `isEcoLTComfort` will now get the correct value.

### Verification Process

I fixed the code manually within WebUI and tabbed in/out of that field.

### Release Notes

Fixing a Template issue to improve field behavior of Energy Saving Temperatur Settings.

### Contributing checklist


theimo1221:
EDIT: Looking into the second bug with the rounding, I found out `ProofAndSetValue` excpects Min/Max to be strings in case of floating number to use the "." as a detection method for floating numbers.
As those where pure numbers within the `hmipChannelConfigDialogs.tcl` I surronded those with quotes resulting in them beeing handled as strings, rather than beeing converted by browser already from `15.0 -> 15` which made the `ProofAndSetValue` Function believe we have an integer field here.

jens-maus (changes requested):
Thanks for the PR. However:

1. Please check the "Files changed" tab for the actual changes you are suggesting. You will notice that your PR also contains changes unrelated to your suggested changes. This happens if you submit PRs based on the master branch which should always be in line and updated with upstream (this repo). Thus, you need to rebase/update your PR branch to the latest changes of the RaspberryMatic repo so that these changes will vanish.
2. Please don't mix multiple issue into a single WebUI Patch. So please remove c458e2d from this PR and submit this change in a separate PR with a new WebUI patch number assigned. Please note, however, that I highly doubt that surrounding in " will be enough here. Please note that this `ProofAndSetValue` function is highly complex and that some devices provide their values as float and some as integers and it doesn't actually suprise me why eQ3 has so much trouble in getting it properly fixed. Nevertheless, if you feel your suggestions are correctly and fix something, please bring them up in a separate PR.

theimo1221:
> 1. Please check the "Files changed" tab for the actual changes you are suggesting. You will notice that your PR also contains changes unrelated to your suggested changes. This happens if you submit PRs based on the master branch which should always be in line and updated with upstream (this repo). Thus, you need to rebase/update your PR branch to the latest changes of the RaspberryMatic repo so that these changes will vanish.

I'm sorry that's absolutely correct, but even after the rebase I always get some changes for Patch 0019, which I have to exclude.

> 2. Please don't mix multiple issue into a single WebUI Patch
"All theses" changes are in the same location regarding the issues written in #1058, which are both the change to `NaN` due to invalid HTML in the template and the missing hidden input with the old value (parallel to the other field). As he can't find the field he can't process the value.
Additionally right in the corresponding lines there are the calls to 'ProofAndSetValue' without quoted numbers:
image
As seen in the screen, by default a passed number ending with zero get's trimmed by the browser.

Same happens here:
image
If we have Min/Max passed as a number, we immediately end in the section where the input value get's rounded (marked in yellow).
This occured while debugging it:
image
And the parent call from the previously yellow marked stack entry:
image
In this last screenshot you see the call beeing made with real numbers, which loose the ".0" as presented in the first screenshot.


So IMO this patch while fixing both errors presented in #1058 is a fix for the same topic.
But sure, if you insist I'll split it in separate ones 👍

theimo1221:
I moved the "separate" fix to #1142. So feel free to review them separately

jens-maus (approved):

jens-maus:
Thanks for this nice PR. After a quick check I merged it accordingly.

> Edit: Additionally I gave you contributor rights to my fork, so in case of minor necessary changes feel free to perform them.

Full contributor rights are not necessary. All you have to do is make sure you only submit a PR on a dedicated feature branch and not on the master/main branch because for feature branches the maintainers of the repositories you send the PR to automatically get permissions to add changes to this branch. However, this is not the case for the master/main branch, that's why it didn't work for your first PR from yesterday.

PR #1142, opened 2021-02-19
0113 - Separate Fixem from 0112 in 2 Patches

theimo1221:
### Description
This is the separated fix as discussed in https://github.com/jens-maus/RaspberryMatic/pull/1139.
This fix surronds the `Min/Max` Values in calls to the `ProofAndSetValue` by quotes to keep the "." in case of number ending with `.0`.  A more detailed description with images can be found in #1139.

### Related Issue

This fixes parts of #1058 as there is written some values like `23.5` get rounded to `23`

### Types of changes

### Alternate Designs

I thought about changing the `PrrofAndSetValue` Function first, but thought to reduce any possible sideeffects, it's better looking into the issue, why ProofAndSetValue gets unexpected values, leading to the faulty rounding.

### Possible Drawbacks

Now, some more fields are interpreted as floating number fields, thus accepting e.g. `23.5` if either the minimum or the maximum contains a `.x`.

### Verification Process

Surronding the calls numbers in the call with '"' resulted in Browser keeping `Minimum="15.0"` as `"15.0"` allowing `ProofAndSetValue` to detect having a field with allowed floating numbers.

### Release Notes

Fixing some internal calls to allow floating numbers being entered in Energy Saving Temperatur Settings.

### Contributing checklist


jens-maus:
Thanks for another nice Bugfix patch. I only had to slightly fix your patch because you put the tcl file in the wrong directory and I also renamed it to 0113 accordingly.

Lets see what the nightly build testers will say about your fix. Really highly appreciated. And please keep up the pace in submitting new PRs as we really need more talented web developers in this project :)

PR #1535, opened 2021-11-16
Updated to latest generic_raw_uart kernel modules

alexreinert:
Updated to latest generic_raw_uart kernel modules


jens-maus:
Thx for the proposed PR updating the generic_raw_uart driver. As I can see that this update will introduce cryptographic device checks, does that mean that a firmware update of a HB-RF-USB or HB-RF-USB-2 is required before using the new version? And what about the non-commercial HB-RF-USB clones from the community? Are they still supported?

alexreinert:
You should know me better. ;-)
Right now unverified devices using the official schematic are still supported, the verification state is just printed to dmesg. As stated in the Homematic Forum I will provide a solution to add signatures to existing commercial devices and for existing and future community based devices. 
 of course I will not activate a hard verification before that and in that case I will  communicate about that breaking change.

jens-maus:
Thanks for clarification! Perfect solution indeed.

PR #1550, opened 2021-11-20
0153-WebUI-dynamic-resizing

ptweety:
### Description

Remove dynamic resizing from the WebUI. This PR is meant to be a starting point for this kind of improvements. Low hanging fruits are e.g. login and logout pages.

### Related Issue

See discussion [here](https://homematic-forum.de/forum/viewtopic.php?f=65&t=41974&start=800#p688952)

### Types of changes

### Verification Process

- [x] Spot checks it in my local dev environment

### Release Notes

- login.htm and logout.htm now resize without extra javascript function calls
- reworked `webui.js` to move from `setStyle` function calls over to CSS `styles.css`

### Contributing checklist


buildroot-external/patches/occu/0021-WebUI-Remove-Logo/occu/WebUI/www/rega/login.htm
```diff
@@ -69,7 +69,7 @@
         <td id="navbar">
           <table cellpadding="0" cellspacing="0">
             <tr>
-              <td style="text-align: center;"><div style="width: 180px;"><img src="/ise/img/homematic_logo_small.png" alt="HomeMatic Logo" /></div></td>
+              <td style="text-align: center;"><div style="width: 180px;"><img src="/ise/img/rm-logo_small.png" alt="RaspberryMatic Logo" /></div></td>
```
jens-maus, line 72: These and the following changes to `login.htm` and `logout.htm` files in the other WebUI patchsets seem to be unnecessary. A usual better approach would be to always use the original files from the occu github repository as the base files.

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,126 @@
...
+-
+-        $("navbar").setStyle({height: wHeaderHeight + "px"});
+-        $("content").setStyle({height: wContentHeight + "px"});
+-        $("footer").setStyle({height: wFooterHeight + "px"});
+-      }
+     </script>
+   </head>
+   <body>
+-    <table style="width:100%; height:100%;" cellpadding="0" cellspacing="0" id="maintable">
++    <table style="width:100vw; height:100vh;" cellpadding="0" cellspacing="0" id="maintable">
```
jens-maus, line 32: If we on changing the style anyway. Why not remote these element-specific `style=` definitions here alltogether and better put them into the global `style.css`?

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,126 @@
...
+       <tr>
+         <td id="navbar">
+           <table cellpadding="0" cellspacing="0">
+@@ -88,7 +67,7 @@
+       </tr>
+       <tr>
+         <td>
+-          <div id="content" style="background-repeat:no-repeat;background-position:480px center;">
++          <div id="content" style="background-repeat:no-repeat;background-position:480px center;height:100%;">
```
jens-maus, line 41: same here

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,126 @@
...
+-
+-        $("navbar").setStyle({height: wHeaderHeight + "px"});
+-        $("content").setStyle({height: wContentHeight + "px"});
+-        $("footer").setStyle({height: wFooterHeight + "px"});
+-      };
+-    </script>
+   </head>
+   <body>
+-    <table style="width: 100%; height: 100%;" cellpadding="0" cellspacing="0" id="maintable" class="j_translate">
++    <table style="width: 100vw; height: 100vh;" cellpadding="0" cellspacing="0" id="maintable" class="j_translate">
```
jens-maus, line 85: same here

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,126 @@
...
+       <tr>
+         <td id="navbar">
+           <table cellpadding="0" cellspacing="0">
+@@ -47,7 +26,7 @@
+       </tr>
+       <tr>
+         <td>
+-          <div id="content" style="background-repeat:no-repeat;background-position:480px center;">
++          <div id="content" style="background-repeat:no-repeat;background-position:480px center;height:100%;">
```
jens-maus, line 94: same here

jens-maus (changes requested):

buildroot-external/patches/occu/0021-WebUI-Remove-Logo/occu/WebUI/www/rega/login.htm
ptweety: I have reverted that and squashed all commits into a new one. Please check

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: Well, one could introduce a new tag `#maintable` in `style.css` and do all the styling from there. Let me think ...

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: fixed

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: fixed

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: fixed

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,142 @@
...
+-
+-        $("navbar").setStyle({height: wHeaderHeight + "px"});
+-        $("content").setStyle({height: wContentHeight + "px"});
+-        $("footer").setStyle({height: wFooterHeight + "px"});
+-      }
+     </script>
+   </head>
+   <body>
+-    <table style="width:100%; height:100%;" cellpadding="0" cellspacing="0" id="maintable">
++    <table cellpadding="0" cellspacing="0" id="maintable">
```
jens-maus, line 32: Any chance to get rid of this old-style `cellspacing` and `cellpadding` and replacing them by CSS statements?

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,142 @@
...
+       gotoLoginPage = function() {
+         location.href = "/login.htm";
+       };
+--- occu/WebUI/www/webui/style.css.orig
++++ occu/WebUI/www/webui/style.css
+@@ -134,9 +134,23 @@
+  * * *   Element IDs                     * * *
+  * * * * * * * * * * * * * * * * * * * * * * */
+ 
++/* Login- und Logout-Fenster */
```
jens-maus, line 114: Please english language comments only!

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,142 @@
...
++  height: 100vh;
++}
++
++#maintable #content {
++  height: 100%;
++  background-repeat: no-repeat;
++  background-position: 480px center;
++}
++
+ /* Navigation-Leiste */
```
jens-maus, line 127: same here

jens-maus (changes requested):

jens-maus:
@ptweety Ok, I have tested your proposed WebUI patch now and to me this looks good so far. The only missing things for the current state of affairs I have commented/reviewed above. Please revise.

In addition, would you like to add your main `webui.js` modifications here right away? Because it would be great to have these major modifications to improve the dynamic resizing of the WebUI using CSS methods in the same PR.

ptweety:
@jens-maus Sure, I'm still working on the patch and will come up with some more commits soonish.

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: It could be done, but I'm unsure if it should be in scope of this PR. I would prefer to have this fix in another one.

jens-maus: Applying subsequent changes to the same line in multiple patches can be quite critical and introduce conflicts in applying patches after another. Thus this would introduce hard dependencies between webui patches and don't make them independent anymore.

ptweety: I just thought about another PR on this patch. There is only so much one can do in a day 😊

jens-maus: IMHO it would be better to integrate this change also in this PR since it is also changing the style stuff. Or you integrate this into the generic `0039-WebUI-Fix-Style-Glitches` which tries to improve the general styling stuff on the WebUI.

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: Fixed

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: fixed, but `/* Navigation-Leiste */` is an original comment.

jens-maus: You are right! then it is fine to keep it!

ptweety:
Ok, patch has been renamed and improved. Please also check the updates made to 0039-patchset.

What I have not replicated from the `onResize()` function into css  is a check against `min-width` and `min-height` of the WebUI. Is this a must or should we leave it?

jens-maus:
> Ok, patch has been renamed and improved. Please also check the updates made to 0039-patchset.

To me they are looking fine. However, what's with the mentioned removal of `cellpadding` and `cellspacing`?

> What I have not replicated from the `onResize()` function into css is a check against `min-width` and `min-height` of the WebUI. Is this a must or should we leave it?

Which check do you mean exactly? Please point your finger at it ;)

buildroot-external/patches/occu/0153-WebUI-DynamicResizing.patch
ptweety: should now also be fixed with f98f231

ptweety:
> To me they are looking fine. However, what's with the mentioned removal of `cellpadding` and `cellspacing`?

It's included now.

> > What I have not replicated from the `onResize()` function into css is a check against `min-width` and `min-height` of the WebUI. Is this a must or should we leave it?
> 
> Which check do you mean exactly? Please point your finger at it ;)

See here:
```
  MIN_WIDTH: 750,  
  MIN_HEIGHT: 400,

...

  if (width  < this.MIN_WIDTH)  { width  = this.MIN_WIDTH;  bodyOverflow = "auto"; }
  if (height < this.MIN_HEIGHT) { height = this.MIN_HEIGHT; bodyOverflow = "auto"; }
```

We could set these minimum values also in the css. But I wonder why.

jens-maus:
> It's included now.

Perhaps! Thanks.

> > Which check do you mean exactly? Please point your finger at it ;)
> 
> See here:
> 
> ```
>   MIN_WIDTH: 750,  
>   MIN_HEIGHT: 400,
> 
> ...
> 
>   if (width  < this.MIN_WIDTH)  { width  = this.MIN_WIDTH;  bodyOverflow = "auto"; }
>   if (height < this.MIN_HEIGHT) { height = this.MIN_HEIGHT; bodyOverflow = "auto"; }
> ```
> 
> We could set these minimum values also in the css. But I wonder why.

I dunno. I would say, remove them for the time being and finish your dynamic resizing changes in this PR so that we can test it altogether. My guess is, that the original eQ3 author thought about there have to be a minimum size so that WebUI elements won't start overlapping.

ptweety:
> I dunno. I would say, remove them for the time being and finish your dynamic resizing changes in this PR so that we can test it altogether.

This PR should now be done.

jens-maus:
> > I dunno. I would say, remove them for the time being and finish your dynamic resizing changes in this PR so that we can test it altogether.
> 
> This PR should now be done.

Ok, I will have a deeper look at it tomorrow. But what about all these `FIX:` and commented out things in the patch?

buildroot-external/patches/occu/0039-WebUI-Fix-Style-Glitches/occu/WebUI/www/webui/style.css
```diff
@@ -184,6 +184,11 @@ a img {
   display: contents;
 }
 
+#divDL, #divRTD, #divFL, #divHTF, #divRL, #divHTR
+{
+  height: calc(100% - 20px); /* FIX: container - 2 * margin */
```
jens-maus, line 189: FIX: comment, please revise.

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,282 @@
...
+ }
+ 
+--- occu/WebUI/www/webui/webui.js.orig
++++ occu/WebUI/www/webui/webui.js
+@@ -6942,6 +6942,8 @@
+    */
+   onResize: function()
+   {
++    /**
++     * not required any more, since we moved to css styles
```
jens-maus, line 241: If really not required, please remove completly! And any chance to get the whole `onResize()` function and the event handler binding eliminated as well?

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,282 @@
...
++    */
+ 
+     if (this.currentPage) { this.currentPage.resize(); }
+ 
+@@ -20235,6 +20238,8 @@
+    **/
+   resizeUser: function()
+   {
++    /**
++     * not required any more, since we moved to css styles
```
jens-maus, line 258: No stray comments, if not required please remove it altogether. And best would be to get rid of the whole `resizeUser()` function if possible.

jens-maus (changes requested):

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
ptweety: done, see 9a27326

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
ptweety: done, see 9a27326

buildroot-external/patches/occu/0039-WebUI-Fix-Style-Glitches/occu/WebUI/www/webui/style.css
ptweety: done, see 84cda45

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,333 @@
...
+-    if (height < this.MIN_HEIGHT) { height = this.MIN_HEIGHT; bodyOverflow = "auto"; }
+-    var contentHeight = height - this.STATIC_HEIGHT;
+-
+-    if ($("body"))    { Element.setStyle("body", {"overflow": bodyOverflow, "width": width  + "px", "height": height + "px"}); }
+-    if ($("header"))  { Element.setStyle("header" , {"height": this.HEADER_HEIGHT  + "px", "width": width + "px"}); }
+-    if ($("menubar")) { Element.setStyle("menubar", {"height": this.MENUBAR_HEIGHT + "px", "width": width + "px"}); }
+-    if ($("content")) { Element.setStyle("content", {"height": contentHeight       + "px", "width": width + "px"}); }
+-    if ($("footer"))  { Element.setStyle("footer" , {"height": this.FOOTER_HEIGHT  + "px", "width": width + "px"}); }
+-
+     if (this.currentPage) { this.currentPage.resize(); }
```
jens-maus, line 254: Any chance to get this also removed/replaced so that the whole `onResize()` function can be removed?

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
```diff
@@ -0,0 +1,333 @@
...
+@@ -20493,11 +20430,7 @@
+     }
+   },
+   
+-  /**
+-   * Berechnet die Gr��e der Startseite f�r den Gast
+-   **/
+-  resizeGuest: function() { }   
+-  
++  resize: function() { }
```
jens-maus, line 330: Why is this dummy `resize()` function still required?

jens-maus (changes requested):

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
ptweety: I had bad luck removing this. My WebUI was not displaying pages correctly.

buildroot-external/patches/occu/0154-WebUI-DynamicResizing.patch
ptweety: Same as above: I had bad luck removing this. My WebUI was not displaying pages correctly.

ptweety: I also stumbled upon an error that was generated by this call:

```
  /**
   * Lädt eine Seite.
   */
  enter: function(page, options)
  {
    Debug.assert(Interface.isImplemented(page, IPage), "WebUI.enter: Interface IPage not implemented");
    
    if (this.currentPage) { this.currentPage.leave(); }
    
    this.previousPage        = this.currentPage;
    this.previousPageOptions = this.currentPageOptions;
```

jens-maus:
@ptweety Ok, I have tested your final PR now and was even able to remove the rest of the `onResize()` and `resize()` functions in `webui.js` so that we show have complete dynamic resizing now. However, have you had a look also at the `m_resize()` functionality in `webui.js`? Where no dynamic resizing is still used for some dialogs. Perhaps you can rework this part in `webui.js` as well.

jens-maus (approved):

ptweety:
> @ptweety Ok, I have tested your final PR now and was even able to remove the rest of the `onResize()` and `resize()` functions in `webui.js` so that we show have complete dynamic resizing now.

Nice 👍 

> However, have you had a look also at the `m_resize()` functionality in `webui.js`? Where no dynamic resizing is still used for some dialogs. Perhaps you can rework this part in `webui.js` as well.

Yes, I had a look. There is almost everything hardcoded - all the sizes of the elements in the dialogs (including margin, padding, borders, ...) and then there are some sprinkles of CSS ... I ... looked away.

jens-maus:
> > However, have you had a look also at the `m_resize()` functionality in `webui.js`? Where no dynamic resizing is still used for some dialogs. Perhaps you can rework this part in `webui.js` as well.
> 
> Yes, I had a look. There is almost everything hardcoded - all the sizes of the elements in the dialogs (including margin, padding, borders, ...) and then there are some sprinkles of CSS ... I ... looked away.

Ok, would be great to have that replaced by pure CSS as well if possible.

ptweety:
> Ok, would be great to have that replaced by pure CSS as well if possible.

It is possible, of course, but it is a huge effort.

Let me give you a simple example by analyzing the `MessageBox`. To paint this simple component we call the `.show()` function of this `class` with some text, a given `width` and `height` and some other parameters (which may influence the layout and sizing as well). It consists of:

- a `UIFrame` derived from `UIContainer` (which is derived from `UIComponent`)
  - a `UIFrameTitle`
  - a `UIFrameContent`
    - a `UIText`
    - a `UIButton`
      - a `UIButtonText`

all of which have `position: absolute` and individually calculated `top`, `left`, `width` and `height`. The later happens not only in one place. It' spread around all of the different components and obviously also in the constructor (initialize) and other functions of the `MessageBox` as well.

If we now want to at least get rid of the calculated `top` and `left` attributes we can introduce a new function
`setPosition` into the `UIFrame` to overload the original function from `UIComponent` as such:

```
  setPosition: function() {
    Element.setStyle(this.m_element, {
      //top : parseInt(y) + "px",
      //left: parseInt(x) + "px"
      top      : "50%",
      left     : "50%",
      transform: "translate(-50%, -50%)"
    });
    return this;
  },
```

That fixes one thing only.

But wait ... the `MessageBox` is also draggable. Guess what happens when a user grabs the `UITitle` of this `MessageBox`? Yes, all of the above was for nothing. We have again dynamically calculated `top` and `left`.

BTW: I wonder why the hell are these components draggable in the first place? I don't understand this feature ...

As a helpful reading regarding modal dialogs let me reference [this article](https://css-tricks.com/considerations-styling-modal/)

jens-maus:
> > Ok, would be great to have that replaced by pure CSS as well if possible.
> 
> It is possible, of course, but it is a huge effort.

Ok, I understand. Then I think we are done here as my own tests look quite good here and you at least managed to get rid of this strange manual resizing chain in `onResize()` which also puzzled me a while ago. Now this looks way cleaner and AFAICS also speeded up resize operations a bit in the WebUI. So I think it is time to merge this on the master and get some feedback from the snapshot testers around here. Thanks again and please continue sending in such great PRs for improving the WebUI even on a level that is not immediately visible to the end user!

Baxxy13:
It looks good with Chromium based Browsers. 
But with Firefox (94.0.2 (64-Bit) ) there are resizing problems on every page except "Einstellungen > Systemsteuerung".
Examples:
RM_Startseite_reseize_firefox
RM_Sysvars_reseize_firefox

jens-maus:
@Baxxy13 Please see if adding `width: 100%` to `.Layer0` in the CSS also solves these issues for you?

Bildschirmfoto 2021-11-24 um 09 35 07

Baxxy13:
after adding `width: 100%` to `.Layer0` everything looks correct now

ptweety:
@Baxxy13 : thanks for testing. `login.htm` and `logout.htm` also need some fix for Firefox browser

jens-maus:
> after adding `width: 100%` to `.Layer0` everything looks correct now

Unfortunately I have to use `width: -moz-available` in `.Layer0` instead because otherweise Safari has horizontal resizing issues of large tables, as it seems. @ptweety perhaps you can test all this throughout all major browsers again? would be great.

ptweety:
@jens-maus: I checked with Safari, Chrome and Firefox (no Edge here). Your fix seems to work fine.

However, for `login.htm` and `logout.htm` we also need to extend the existing style as follows:

```
#maintable #content {
  height: 100%;
  width: 100%;
  display: inline-table;
  background-repeat: no-repeat;
  background-position: 480px center;
}
```

Baxxy13:
> `login.htm` and `logout.htm` also need some fix for Firefox browser

i saw that too, therefore i wrote...


> on _every_ page except...

RM_login_reseize_firefox

jens-maus:
> > `login.htm` and `logout.htm` also need some fix for Firefox browser
> 
> i saw that too, therefore i wrote...
> 
> 
> > on _every_ page except...

That was to unspecific ;)

ptweety:
@jens-maus will you make a commit with the style updates for `login.htm` and `logout.htm` or shall I create a PR?

ptweety:
Ah, forget it. I was to late to the party (outdated browser window...)

andrzejgilas:
scrolling the WebUI page doesn't seem to work properly either, the right side doesn't fit

Animation

jens-maus:
> scrolling the WebUI page doesn't seem to work properly either, the right side doesn't fit

This seems to be yet-another-firefox only problem as it seems. Neither Chrome nor Safari seem to show this behaviour and the `contentLeft` can be independently scrolled from the content in `contentRight`. But perhaps @ptweety has some idea where this comes from or how to fix it for firefox.

ptweety:
Yes, let's start with a small analysis.

The HTML of the WebUI has several sections. One being the `#content`. Many pages directly inject their data inside this container, hence also rely on the behavior of the same. This includes the scrolling (implemented with `overflow: auto`.

But in some cases the container is not filled with a simple table that needs to be scrolled, but instead is again divided in subsections like `contentLeft` or `contentRight` which need individual scrolling. In this cases `#content` should not do any scrolling. As a consequence the container may not have any `overflow` by it's own.

Unfortunately different browsers implement different behavior if it comes to the `overflow` in such sections. Chrome and others reduce the dimensions, but Firefox expands to the max (`height`).

Lets go more visual:
Bildschirmfoto 2021-11-25 um 10 38 46

You can see the differences in the calculated height in both browsers (top: Chrome, bottom: Firefox)

To fix this we maybe need to rework the affected pages and greatly reduce their complexity of the HTML. In particular the deep nesting of tables inside tables which have again tables inside ... needs to be brought back to one simple layer.

ptweety:
Now, let's compare with a working example (`startpage`):
Bildschirmfoto 2021-11-25 um 10 50 52

Here the HTML is much simpler without deep nesting. You may also find that the structure is build with `<div>` instead of `<table>`.

jp112sdl:
Is there a solution in sight?

jens-maus:
> Unfortunately different browsers implement different behavior if it comes to the `overflow` in such sections. Chrome and others reduce the dimensions, but Firefox expands to the max (`height`).

Isn't there a css statement for firefox to prevent it from explicitly using the `#content` container for scrolling and just rely on `contentLeft` and `contentRight`?

> To fix this we maybe need to rework the affected pages and greatly reduce their complexity of the HTML. In particular the deep nesting of tables inside tables which have again tables inside ... needs to be brought back to one simple layer.

I am all for it. However, we have to be careful that this doesn't end up in a complete rewrite of the WebUI which would mean we are completly off to future changes eQ3 might introduce in future versions. And what about our mid-term plan to better push the bootstrap.css integration for a new WebUIv2 development? I really fear we are splitting our resources here by trying to improve the existing WebUI on one hand and trying to get WebUIv2 development going as we already discussed/decided.

ptweety:
Well, I think that the fix is quite simple and does not introduce to much of a change. For `devices` the idea is something like that:
Bildschirmfoto 2021-11-25 um 11 18 25

But to provide a fix, I need some more time (after work).

jens-maus:
Hmm, ok. If this would only mean to replace this `CLASS03900` table and their `contentLeft` and `contentRight` `<td>` entries with simple `<div>` statements, I am all for it! So please submit a new PR on it ASAP ;-)

ptweety:
OK, here is started to prepare a PR: https://github.com/jens-maus/RaspberryMatic/compare/master...ptweety:fix-dynamic-resizing

I have not yet build a new image to test it locally, but maybe it already shows off what needs to be fixed.

jens-maus:
create a PR out of it and we can work it out that way.

PR #1847, opened 2022-05-24
lighttpd config optimizations

gstrauss:
### Description

lighttpd config optimizations to reduce mem usage, reduce redundancy (ease maintenance), speed up config processing

Please review commits individually from oldest to newest.
Commits *have not* been tested.  Changes proposed are a proof of concept (PoC) for discussion.

### Types of changes

### Possible Drawbacks

Proposed TLS configuration change removes use of older CBC ciphers.  This might possibly affect very old clients with very old TLS libraries supporting only CBC ciphers.

### Verification Process

Note: none; PoC

### Release Notes

- lighttpd config optimizations

### Contributing checklist


gstrauss:
FYI: another possible improvement: depending on how/where the rega backend stores the username/password, mod_authn_rega might be replaced with [lighttpd mod_magnet](https://wiki.lighttpd.net/Docs_ModMagnet) and a small lua script which performs HTTP Basic Auth by reading the file containing the password for 'user'.  Example: [lighttpd lua mod_auth](https://wiki.lighttpd.net/ModMagnetExamples#lua-mod_auth)  The lua script could be enhanced to `stat()` the password file each request and could read and cache the new values when the file is changed.

buildroot-external/overlay/base/etc/lighttpd/conf.d/proxy_normal.conf
```diff
@@ -27,22 +27,20 @@ $SERVER["socket"] == ":8181" {
   proxy.server =  ( "" => ("localhost" => ("host" => "127.0.0.1", "port" => 8183)))
   include "/var/etc/lighttpd_auth.conf"
 }
-$SERVER["socket"] == "[::]:8181" {
-  server.use-ipv6 = "enable"
```
jens-maus, line 89: Is removing `server.use-ipv6` here really non-critical? AFAIR I also had to add that back when I added IPv6 support so that lighttpd really started to work correctly.

buildroot-external/overlay/base/etc/lighttpd/lighttpd.conf
```diff
@@ -131,6 +131,14 @@ server.pid-file = state_dir + "/lighttpd.pid"
...
 #server.max-worker = 4
 
+##
+## server feature flags
+##
+## enable immediate graceful restart; finish existing connections in background
+##
+server.feature-flags += ("server.graceful-restart-bg" => "enable")
+server.systemd-socket-activation = "enable"
+
```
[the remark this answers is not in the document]
gstrauss: `server.systemd-socket-activation` does not require systemd.  The feature allows lighttpd to use LISTEN_FDS and LISTEN_PID environment variables -- on systemd-based systems defined by systemd for socket activation -- to tell the next generation of lighttpd how many listening sockets to find already opened in its new environment.  This is used for immediate graceful restart to be able to continue serving requests with zero downtime (the listen socket was never closed) and a minimal pause to reading and process the lighttpd config while starting up the next lighttpd generation.

buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd
```diff
@@ -82,7 +82,7 @@ reload() {
     echo "include \"/etc/lighttpd/conf.d/proxy_normal.conf\"" >/var/etc/lighttpd_proxy.conf
   fi
 
-  start-stop-daemon -K -s HUP -q -p ${PIDFILE_ANGEL}
+  start-stop-daemon -K -s USR1 -q -p ${PIDFILE_ANGEL}
```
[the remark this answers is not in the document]
gstrauss: Yes, `SIGUSR1` is supported by `lighttpd-angel`.

buildroot-external/overlay/base/etc/lighttpd/conf.d/sslsettings.conf
```diff
@@ -1,9 +1,5 @@
 ssl.engine = "enable"
 ssl.pemfile = "/etc/config/server.pem"
-ssl.honor-cipher-order = "enable"
-ssl.cipher-list = "EECDH+AESGCM:AES256+EECDH"
-#ssl.cipher-list = "EECDH+AESGCM:EDH+AESGCM:AES256+EECDH:AES256+EDH"
-ssl.use-sslv2 = "disable"
-ssl.use-sslv3 = "disable"
-ssl.openssl.ssl-conf-cmd = ("Protocol" => "-TLSv1, -TLSv1.1, -SSLv3")
-ssl.read-ahead = "disable"
```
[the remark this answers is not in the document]
gstrauss: `ssl.read-ahead = "disable"` is the default since lighttpd 1.4.46.

buildroot-external/overlay/base/etc/lighttpd/conf.d/proxy_normal.conf
gstrauss: lighttpd treats addresses that begin with `[` as `AF_INET6`.

buildroot-external/overlay/base/etc/lighttpd/conf.d/proxy_normal.conf
```diff
@@ -18,9 +18,11 @@ else $SERVER["socket"] == "[::]:80" {
 # WebUI HTTPS (TLS)
 #-------------------
 else $SERVER["socket"] == ":443" {
+  ssl.engine = "enable"
```
jens-maus, line 21: Here and on all https ports following I had to add back `ssl.engine = "enable"` because otherwise I receive the following error in `lighttpd-error.log`:

```
2022-05-25 09:17:09: (../src/mod_openssl.c.3295) SSL: error:140BA0C3:SSL routines:SSL_new:null ssl ctx
```

So, somehow your global ssl.engine enable and then just disabling for port 80 / non-encrypted sites does not seem to work. After some investigation it seems this is similar to https://redmine.lighttpd.net/boards/2/topics/9620, I guess.

buildroot-external/overlay/base/etc/lighttpd/conf.d/proxy_normal.conf
```diff
@@ -26,23 +26,17 @@ else $SERVER["socket"] == "[::]:443" {
   include "/etc/lighttpd/conf.d/webui.conf"
 }
 
-#-------------------------------------
-# else config auth for everything else
-#-------------------------------------
-else {
-  include "/var/etc/lighttpd_auth.conf"
```
jens-maus: Here it is similar to the ssl.engine thing. I had to add back the include to every `$SERVER` definition and remove it here from the global space, because as soon as I enabled authentication, even port 80 and 443 was requesting basic authentication, which is not the wanted behaviour. Only these other ports like 8181 should get an authentication enabled, but port 80 and 443 should never be forced to an authentication.

buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd
```diff
@@ -82,7 +82,7 @@ reload() {
     echo "include \"/etc/lighttpd/conf.d/proxy_normal.conf\"" >/var/etc/lighttpd_proxy.conf
   fi
 
-  start-stop-daemon -K -s USR1 -q -p ${PIDFILE_ANGEL}
+  start-stop-daemon -K -s HUP -q -p ${PIDFILE_ANGEL}
```
jens-maus, line 85: I had to revert back to sending `HUP` instead of just `USR1` because when turning on/off the authentication or https redirection lighttpd seem to not have refreshed all things correctly and returned some permission denied or `403 Forbidden` even on port 443. Changing the reload() method to use HUP instead of USR1 immediately solved this issue.

jens-maus (commented):
@gstrauss Please have a look at my recent changes/fixes to your proposed PR here. After having tested your changes, some of them seem to not work as expected and required some tuning/fixing on my side. Perhaps you have some further thoughts on these reverts/changes. which could lead to some other suggestions from you how to tune these configs even further.

gstrauss:
Strange.  I need to look into those issues further.  Thank you for testing.

Sending SIGHUP to lighttpd-angel turns around and sends SIGINT to lighttpd, resulting in a graceful exit, after which lighttpd-angel restarts lighttpd.  Sending a SIGUSR1 instead results in a graceful restart of lighttpd.  I believe that works as intended, but think there might be an issue with some of the other changes.

gstrauss:
I have not yet been able to duplicate the behavior you reported regarding auth.  Did it occur for you only after restarting lighttpd?  Did it occur for you only with SIGUSR1?  Does the issue occur if you comment out: `server.feature-flags += ("server.graceful-restart-bg" => "enable")` in lighttpd.conf?

jens-maus:
> I have not yet been able to duplicate the behavior you reported regarding auth. Did it occur for you only after restarting lighttpd? Did it occur for you only with SIGUSR1? Does the issue occur if you comment out: `server.feature-flags += ("server.graceful-restart-bg" => "enable")` in lighttpd.conf?

You are right. I just retested it again and the issue really just happens with using `SIGUSR1` to reload the config. As soon as HUP is used everything is refreshed correctly with no issue appearing for port 80/443 anymore. Therefore I reverted my change now with commit d76e023.

buildroot-external/overlay/base/etc/lighttpd/conf.d/sslsettings.conf
```diff
@@ -1,9 +1,6 @@
+server.port = 443
```
jens-maus, line 1: Is this recent addition really right? because we use ssl not only for port 443 but also for all the 4xxxx ports in `proxy_normal.conf`.

gstrauss: Yes, added.  If enabling `ssl.engine = "enable"` in global scope, I wanted it to apply to port 443, not the default (`server.port = 80`).  All the `$SERVER["socket"]` which set `ssl.engine = "enable"` will inherit the global TLS config if they do not set their own `ssl.*` directives besides `ssl.engine = "enable"`.

gstrauss:
I did not realize you were in the process of looking at this and I just pushed a slight alternative auth config.  Both this and the prior slightly reduce memory use by having the regex for this config snippet once rather than 26 times.  At runtime, the difference during configuration processing for requests on port 80 and 443 is one extra regex, but 26 fewer string compares with other sockets.

gstrauss:
Regarding SIGUSR1, does the issue occur if you comment out: `server.feature-flags += ("server.graceful-restart-bg" => "enable")` in lighttpd.conf and use SIGUSR1?

jens-maus:
> Regarding SIGUSR1, does the issue occur if you comment out: `server.feature-flags += ("server.graceful-restart-bg" => "enable")` in lighttpd.conf and use SIGUSR1?

This indeed changes the game. With `server.graceful-restart-bg` removed enabling/disabling auth works. however, when turning https redirection on/off I then still end up with:

```
2022-05-26 00:49:32: (../src/mod_openssl.c.3295) SSL: error:140BA0C3:SSL routines:SSL_new:null ssl ctx
```

So, SIGHUP seems to be still required.

jens-maus:
> I did not realize you were in the process of looking at this and I just pushed a slight alternative auth config. 

This change seems to work fine in my quick tests. So, the auth issue seems to be solved now. The only things missing/to be clarified are

1. SIGHUP still required (even without graceful-restart-bg).
2. This `server.port=443` setting in sslsettings.conf I still do not fully understand. However, it seems to work for the $SERVER definitions even for ports != 443. So why is it required and why does it not have any negative effect?!?

gstrauss:
> SIGHUP still required (even without graceful-restart-bg).

That is still a mystery to me, too.  The SIGHUP to lighttpd-angel turns into a SIGINT to lighttpd, which is a graceful stop.  SIGUSR1 to lighttpd-angel is a SIGUSR1 to lighttpd, which is a graceful restart.  My next area of inquiry will be to trace the scripts which signal lighttpd.  a SIGUSR1 to lighttpd can be very, very fast and if the scripts signalling are making changes to lighttpd config files, there might be a race condition.

> This server.port=443 setting in sslsettings.conf I still do not fully understand. However, it seems to work for the $SERVER definitions even for ports != 443. So why is it required and why does it not have any negative effect?!?

lighttpd uses `server.bind` and `server.port` to configure itself, and uses defaults if those are not specified.  (default is *:80)  This occurs regardless the existence of any `$SERVER["socket"]` which may cause lighttpd to listen on additional addresses/ports.

Since this PR uses `ssl.engine = "enable"` in the global scope, I chose to apply to that to `server.port = 443` where that setting is desirable.  The additional `$SERVER["socket"]` containing only `ssl.*` directive `ssl.engine = "enable"` inherit and reuse the *openssl* `SSL_CTX`, thereby saving memory, instead of having `SSL_CTX` per `$SERVER["socket"]`.

gstrauss:
Hmmm.   Might there be some interplay with files in the overlay versus the config files the existing lighttpd process sees (if gracefully restarting) versus the existing lighttpd process exiting and a newly spawned lighttpd process reading overlaid filesystem?

gstrauss:
I think I found something that might be a bug in lighttpd graceful restart.  lighttpd graceful restart performs full reloads of modules and resets static data *except* for the `server_socket` structures.  (One reason for this is that lighttpd might not be able to reopen some sockets if lighttpd dropped privileges from root.)  However, there is a piece of the `server_socket` struct which needs to be re-initialized if the configuration added or removed lighttpd.conf conditions, and I do not think the data is being reinitialized.

I have pushed an additional commit which simplifies the http to https redirect config.  When you have a chance, please see if that makes a difference with graceful restart (SIGUSR1)

gstrauss:
I have been able to reproduce the error you are seeing with reconfiguration and graceful restart (SIGUSR1).

I have removed the commit which uses SIGUSR1 (and have removed your revert) so that this PR can proceed without it.

I'll submit a separate PR for the graceful restart once I track down and fix the issue.

gstrauss:
FYI: If the rega password for 'user' does not change frequently, but you want it to take effect immediately when it does, then you might want to trigger a lighttpd restart when the password changes.

gstrauss:
FYI: https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/b1e10aea6856bdc7b02ac8a0736dd60a0616635a fixes the issue with graceful restart and reinitializing internal indexes of previously opened listening sockets.  The commit in my dev branch implements for graceful sockets similar re-init steps that were already being done for sockets inherited (and discovered via systemd socket activation environment variables).  The fix will be part of the next lighttpd release.

This PR no longer contains the SIGUSR1 commit, and is not affected.

jens-maus:
> FYI: https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/b1e10aea6856bdc7b02ac8a0736dd60a0616635a fixes the issue with graceful restart and reinitializing internal indexes of previously opened listening sockets. 

Ok, with commit 02556a4 I have now integrated your graceful SIGUSR1 patch as an upstream patch against lighttpd 1.4.64 and after some testing it seems to work now as expected. So I can change the auth and https redirect settings now in the webui and lighttpd reloaded with SIGUSR1 is picking up the changes correctly.

jens-maus:
> FYI: If the rega password for 'user' does not change frequently, but you want it to take effect immediately when it does, then you might want to trigger a lighttpd restart when the password changes.

BTW: Is a SIGUSR1 restart enough for lighttpd to refresh the auth cache or does it require a full lighttpd restart?

gstrauss:
In e70533ed2d06064aace4ee01706d3c0baf4e9f55 you added `"url-invalid-utf8-reject" => "disable"` to buildroot-external/overlay/base/etc/lighttpd/lighttpd.conf but not to buildroot-external/package/recovery-system/external/overlay/base/etc/lighttpd/lighttpd.conf

Thank you for your additional feedback on the `"url-invalid-utf8-reject" => "disable"` patch, and the graceful restart patch.  I plan to release lighttpd 1.4.65 some time in June and containing those patches.

gstrauss:
One reverted part of the original patchset was immediate graceful restart.  With the following, you would no longer need the `sleep 2` in buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd

```
##
## server feature flags
##
## enable immediate graceful restart; finish existing connections in background
##
server.feature-flags += ("server.graceful-restart-bg" => "enable")
server.systemd-socket-activation = "enable"
```

Do you want me to submit an additional patch to this pathset with those changes?

gstrauss:
Minor additional consideration: `sendfile` is very efficient and well-supported in lighttpd, even on many embedded filesystems.  Also, lighttpd falls back to using `writev` if `sendfile` fails.  lighttpd.conf should use lighttpd defaults (which will prefer `sendfile` when available) instead of explicitly specifying `server.network-backend = "writev"` unless there is a specific reason (other than historical).  Allowing lighttpd to use `sendfile` will speed up non-TLS file access, e.g. non-TLS proxying which occurs on localhost for RaspberryMatic.

Combined patch with comment above:
```
diff --git a/buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd b/buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd
index 77b74b3a..a90e68cb 100755
--- a/buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd
+++ b/buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd
@@ -86,7 +83,6 @@ reload() {
   fi

   start-stop-daemon -K -s USR1 -q -p ${PIDFILE_ANGEL}
-  sleep 2
   [[ -f ${PIDFILE} ]] && echo "OK" || echo "ERROR"
 }

diff --git a/buildroot-external/overlay/base/etc/lighttpd/lighttpd.conf b/buildroot-external/overlay/base/etc/lighttpd/lighttpd.conf
index bf804270..7fe51cfd 100644
--- a/buildroot-external/overlay/base/etc/lighttpd/lighttpd.conf
+++ b/buildroot-external/overlay/base/etc/lighttpd/lighttpd.conf
@@ -131,6 +131,14 @@ server.pid-file = state_dir + "/lighttpd.pid"
 ##
 #server.max-worker = 4

+##
+## server feature flags
+##
+## enable immediate graceful restart; finish existing connections in background
+##
+server.feature-flags += ("server.graceful-restart-bg" => "enable")
+server.systemd-socket-activation = "enable"
+
 ##
 #######################################################################

@@ -201,7 +193,7 @@ include "conf.d/debug.conf"
 ## linux-sendfile - is recommended for small files.
 ## writev         - is recommended for sending many large files
 ##
-server.network-backend = "writev"
+#server.network-backend = "writev"

 ##
 ## As lighttpd is a single-threaded server, its main resource limit is
```

jens-maus:
> One reverted part of the original patchset was immediate graceful restart. With the following, you would no longer need the `sleep 2` in buildroot-external/overlay/WebUI/etc/init.d/S50lighttpd
> 
> ```
>[...]
> server.feature-flags += ("server.graceful-restart-bg" => "enable")
> server.systemd-socket-activation = "enable"
> ```

Oh, I didn't recognize that the `server.graceful-restart-bg` flag was not present anymore. Thanks for the hint. However, after having activated it I am back to seeing issues when restarting lighttpd using SIGUSR1 :(

In fact, when enabling/disabling authentication via the WebUI access to port 80 is returning the following curl return:

```
$ curl -v -k http://192.168.5.56
*   Trying 192.168.5.56:80...
* TCP_NODELAY set
* Connected to 192.168.5.56 (192.168.5.56) port 80 (#0)
> GET / HTTP/1.1
> Host: 192.168.5.56
> User-Agent: curl/7.68.0
> Accept: */*
> 
* Empty reply from server
* Connection #0 to host 192.168.5.56 left intact
curl: (52) Empty reply from server
```

and looking into lighttpd-error.log reveals the following:

```
2022-05-30 21:25:03: (../src/mod_openssl.c.3249) SSL: 1 error:1408F09C:SSL routines:ssl3_get_record:http request (192.168.5.2)
```

And when I try to access port 443 I  get again a `401 Unauthorized` return:

```
$ curl -k https://192.168.5.56
<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>401 Unauthorized</title>
 </head>
 <body>
  <h1>401 Unauthorized</h1>
 </body>
</html>
```

However, Port 443 should never get setup for basic authentication like, e.g. port 8181 according to the config.  Removing the whole `server.graceful-restart-bg` from `server.feature-flags` solves the issue again and a USR1 restart works as expected.

So something still does not seem to work correctly with the graceful-restart-bg option in combination with SIGUSR1, I am afraid.

gstrauss:
Hmmm.  I'll see if I can reproduce what you are seeing.  The lighttpd code handling graceful restart is largely the same whether or not `server.graceful-restart-bg` is set.  The main difference is waiting for existing connections to finish (for configured timeout, default 8 seconds) or backgrounding the current lighttpd process (to wait for existing connections to finish) and immediately restarting.

jens-maus:
> I have not been able to reproduce what you are seeing, but I added a few commits to the tip of my dev branch.
> 
> Would you see if they help? https://git.lighttpd.net/lighttpd/lighttpd1.4/src/branch/personal/gstrauss/master

I cherry-picked your commit d08ca831b7 and modified it for 1.4.64 to be applicable cleanly. However, unfortunately it does not solve the issue I have seen lately.

If you cannot reproduce it, would you like to try to reproduce the issue with a running RaspberryMatic installation in place? If so, see if you can install an ova variant of RaspberryMatic which you should be able to quickly startup/use in a VirtualBox or other virtualization environment you might have at hand.

gstrauss:
> If you cannot reproduce it, would you like to try to reproduce the issue with a running RaspberryMatic installation in place?

I'll give that a try later (after some sleep)

Separately, would you mind if I squash some (not all) of the patches in this PR -- including things like whitespace and file renames -- or do you prefer all of the current individual commits?

jens-maus:
> Separately, would you mind if I squash some (not all) of the patches in this PR -- including things like whitespace and file renames -- or do you prefer all of the current individual commits?

No need to squash or force push here anything within this PR because I will anyway squash merge this whole PR once I decide it is ready for a merge.

gstrauss:
> If you cannot reproduce it, would you like to try to reproduce the issue with a running RaspberryMatic installation in place? If so, see if you can install an ova variant of RaspberryMatic which you should be able to quickly startup/use in a VirtualBox or other virtualization environment you might have at hand.

Which keyboard mappings are available in the image?  I see only de.  Unfortunately, I do not have a de keyboard (and am missing keys on my keyboard for the chosen keyboard mapping, so typing is absolutely painful, e.g. missing '>', '<', '|' for typical shell redirection among other things.)  It is not straightforward to set up a working environment for non-Germans.  Is there some documentation you could point me to in order to get started?  I used ssh-keygen to generate missing keys in /usr/local/etc/, and am in the process of seeing why ssh is still not starting...

jens-maus:
Great that you are about to test RaspberryMatic yourself and sorry first if all for the localization issues but RaspberryMatic is most prominent in german speaking countries, that's why defaults are like this traditional ;)

First of all, most of the things even for the console are setup using the WebUI on port 80. So please try to do an initial setup there first. Then you can switch your geo location there to your location because the console keyboard layout is currently bound to the geo location. Or you use setkmap to switch to an English keyboard layout.

And then you can also start/enable  ssh connectivity within the WebUi and set a ssh/root pw there. Then this should bring up everything correctly.

But now that you are going to test the PR changes from here live, I would first need to generate an ova build with these changes and provide that build for you. So please give me some time to prepare that while you are going to do your initial setup in your test VM.

gstrauss:
`setkmap=us` added to grub command line did not have a visible effect for me, and setkmap is not available in the root shell or from busybox.  I am not used to having to change the keyboard layout, so I apologize if there is an obvious way that I am missing.

Using the WebUI, I selected the language as English and there was a prompt about keyboard (and I selected English), but even after rebooting, I still have de keymappings.  Is there some place else in the WebUI that I missed?

jens-maus:
> > But now that you are going to test the PR changes from here live, I would first need to generate an ova build with these changes and provide that build for you. So please give me some time to prepare that while you are going to do your initial setup in your test VM.
> 
> Thank you, but not necessary. If I have ssh, I can put the files I need into place. The hurdle has been slow typing on an unfamiliar (to me) keyboard layout and having to insert unicode sequences for a variety of chars not on my physical keyboard.

It is not only text files that you might have to exchange, but also the lighttpd binary because of the patches I applied to it. So this is really no big work here. So I will prepare an ova build based on this PR here and provide you a download link to testing purposes. Will let you know when done.

> `setkmap=us` added to grub command line did not have a visible effect for me, and setkmap is not available in the root shell or from busybox. I am not used to having to change the keyboard layout, so I apologize if there is an obvious way that I am missing.
> 
> Using the WebUI, I selected the language as English and there was a prompt about keyboard (and I selected English), but even after rebooting, I still have de keymappings. Is there some place else in the WebUI that I missed?

Oh sorry, it is not `setkmap` that you would require and it is not the geolocation that triggers the setup of the de keymap. According to this code passage the selected timezone triggers de keyboard layout being selected as default upon bootup:

```sh
    # if /etc/config/timezone is set to Europe/Berlin we
    # assume a german keyboard layout
    if [[ $(cat /etc/config/timezone) == "Europe/Berlin" ]]; then
      /sbin/loadkmap </etc/de.kmap      
    fi
``` 

However, you can of course also simple use

```sh
/sbin/loadkmap </etc/en.kmap
````

so switch back to english keyboard mapping on the console. Sorry for the hassle :)

**Comment by jens-maus** [comment]
> > > But now that you are going to test the PR changes from here live, I would first need to generate an ova build with these changes and provide that build for you. So please give me some time to prepare that while you are going to do your initial setup in your test VM.
> > 
> > 
> > Thank you, but not necessary. If I have ssh, I can put the files I need into place. The hurdle has been slow typing on an unfamiliar (to me) keyboard layout and having to insert unicode sequences for a variety of chars not on my physical keyboard.
> 
> It is not only text files that you might have to exchange, but also the lighttpd binary because of the patches I applied to it. So this is really no big work here. So I will prepare an ova build based on this PR here and provide you a download link to testing purposes. Will let you know when done.

Ok, the build of this special version is now done. You can download it from here:

https://cloud.light-speed.de/s/8qrtEzi2LEKHtLy

When you already have a working RaspberryMatic you can simply use the WebUI to update to this one by uploading this `*.zip` via `Settings -> Control panel -> CCU Maintenance` and then upload the firmware update there. RaspberryMatic should then take care to update to this version and afterwards you should have an installation with lighttpd patched and the config files setup according to this PR here.

If you decide to modify the lighttpd config make sure to remount the rootfs rw via `mount -o rw,remount /`, then you can modify anything in rootfs (e.g. `/etc/lighttpd/lighttpd.conf`) and then restart lighttpd via `monit restart lighttpd` which is the watchdog taking care of restarting lighttpd, etc.

And to reproduce the issue I was seeing go to the WebUI and then to `Settings -> Control panel -> Security` and then enable/disable the authenticating there as well as the https redirection. Here I can perfectly reproduce the issue by simply turning on the `server.graceful-restart-bg` setting in `lighttpd.conf` and then I simply have to turn on the authentication in the WebUI. afterwards port 80/443 is not accessible anymore (`401 Unauthorized` on port 443 and `ERR_EMPTY_RESPONSE` on port 80). A complete `monit restart lighttpd` solves this issue immediately. And without the `server.graceful-restart-bg` switching on/off authentication does not result in that issue. Hope you can reproduce it yourself.

**Comment by jens-maus** [comment]
> Thank you for the image. After upgrading, I can reproduce that issue you are seeing with `server.graceful-restart-bg` and when enabling auth. I'll try to be creative to get more diagnostics.

Great that you can reproduce it. Let me know if I can be of some further help here. So I guess I will postpone merging this PR until you have provided more info on that matter so that we can perhaps integrate a further improvement here regarding the graceful restart.

**Comment by gstrauss** [comment]
It appears that we slightly crossed paths.  A few days ago, I added to my dev branch https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/d08ca831b7f2fcd7d3b96c01a1ab454c45921048 which re-initializes the `is_ssl` flag after graceful restart.  When you get a chance, would you please add the patch and see if the graceful restart with `server.feature-flags += ("server.graceful-restart-bg" => "enable")` now works as expected?

**Comment by jens-maus** [comment]
> It appears that we slightly crossed paths.

Indeed. It seems this PR gets slightly more complicated over time. Probably time to merge anytime soon and then try to get the graceful restart working in a separate PR...

> A few days ago, I added to my dev branch https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/d08ca831b7f2fcd7d3b96c01a1ab454c45921048 which re-initializes the `is_ssl` flag after graceful restart. When you get a chance, would you please add the patch and see if the graceful restart with `server.feature-flags += ("server.graceful-restart-bg" => "enable")` now works as expected?

Back then (see https://github.com/jens-maus/RaspberryMatic/pull/1847#issuecomment-1141852112)  I tested this commit already after having modified it for a clean lighttpd 1.4.64 because it does not cleanly apply due to other changes around the same lines in your `src/network.c`. However, even after having applied this patch it didn't solve the graceful restart it. Would you like me to add it anyway to this PR and regenerate my test build so that you can download it again and test it accordingly?

**Comment by gstrauss** [comment]
> Indeed. It seems this PR gets slightly more complicated over time. Probably time to merge anytime soon and then try to get the graceful restart working in a separate PR...

lighttpd 1.4.65 will be released soon, so you might want to defer merging this PR until lighttpd 1.4.65 so that there are (many) fewer additional patches to lighttpd.

> However, even after having applied this patch it didn't solve the graceful restart it. 

Since the ova image is x86_64, I built my lighttpd dev branch (including this patch) and installed it into the ova image, where things then appears to work for me.

> Would you like me to add it anyway to this PR and regenerate my test build so that you can download it again and test it accordingly?

Probably worth waiting for lighttpd 1.4.65 to avoid unnecessary work with cherry-picking and merging.  Yes, I agree that the graceful restart using `server.feature-flags += ("server.graceful-restart-bg" => "enable")` can wait for a future PR.

**Comment by jens-maus** [comment]
> lighttpd 1.4.65 will be released soon, so you might want to defer merging this PR until lighttpd 1.4.65 so that there are (many) fewer additional patches to lighttpd.
> 
> > However, even after having applied this patch it didn't solve the graceful restart it.
> 
> Since the ova image is x86_64, I built my lighttpd dev branch (including this patch) and installed it into the ova image, where things then appears to work for me.

This sounds great. If you could provide a 1.4.x snapshot (nothing here yet: http://download.lighttpd.net/lighttpd/snapshots-1.4.x/) of it I could try to integrate it immediately in this PR so that I can create a new build and see if the issue is really solved regarding graceful restart.

**Comment by gstrauss** [comment]
https://git.lighttpd.net/lighttpd/lighttpd1.4/archive/master.tar.gz

(I'll have to look into why snapshots-1.4.x has not been updated in ages)

**Comment by jens-maus** [comment]
> https://git.lighttpd.net/lighttpd/lighttpd1.4/archive/master.tar.gz
> 
> (I'll have to look into why snapshots-1.4.x has not been updated in ages)

Is there also a possibility to fix the download to a certain commit id because otherwise master.tar.gz would always point to the the HEAD of the repo - which of course would change over time when you commit something.

**Comment by gstrauss** [comment]
Generally, what is pushed to lighttpd git master is intended to be stable and (almost) production ready
https://git.lighttpd.net/lighttpd/lighttpd1.4/archive/master.tar.gz is dynamically generated and is fine for a one-off test build.

> Is there also a possibility to fix the download to a certain commit id because otherwise master.tar.gz would always point to the the HEAD of the repo - which of course would change over time when you commit something.

Current HEAD:
https://git.lighttpd.net/lighttpd/lighttpd1.4/archive/ef62ff9419f5742dd5631293a4c19088574ebd99.tar.gz

**Comment by jens-maus** [comment]
Ok, as you can see, I updated lighttpd to 1.4.65 ([ef62ff9419f5742dd5631293a4c19088574ebd99) in this PR. And I could indeed verify that the graceful restart seems to work now. Great!

However, unfortunatly a new problem came up immediately when testing with 1.4.65. Some GET requests to https resources are returning `PROTOCOL_ERROR` kind of errors now:

```
$ curl -vvv -k https://192.168.5.53/webui/css/extern/nfs_datepicker.css
*   Trying 192.168.5.53:443...
* TCP_NODELAY set
* Connected to 192.168.5.53 (192.168.5.53) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: C=DE; O=HomeMatic; OU=1D8997144C; CN=homematic-raspi
*  start date: May 25 07:08:42 2022 GMT
*  expire date: May 22 07:08:42 2032 GMT
*  issuer: C=DE; O=HomeMatic; OU=1D8997144C; CN=homematic-raspi
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x55f262dd00e0)
> GET /webui/css/extern/nfs_datepicker.css HTTP/2
> Host: 192.168.5.53
> user-agent: curl/7.68.0
> accept: */*
> 
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Connection state changed (MAX_CONCURRENT_STREAMS == 8)!
* http2 error: Invalid HTTP header field was received: frame type: 1, stream: 1, name: [], value: [u=2]
* HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
* stopped the pause stream!
* Connection #0 to host 192.168.5.53 left intact
curl: (92) HTTP/2 stream 0 was not closed cleanly: PROTOCOL_ERROR (err 1)
``` 

As you can see, in this request I am just trying to access/download a css content. However, trying to do the same on the main https page proceed cleanly.

```
$ curl -vvv -k https://192.168.5.53/
*   Trying 192.168.5.53:443...
* TCP_NODELAY set
* Connected to 192.168.5.53 (192.168.5.53) port 443 (#0)
* ALPN, offering h2
* ALPN, offering http/1.1
* successfully set certificate verify locations:
*   CAfile: /etc/ssl/certs/ca-certificates.crt
  CApath: /etc/ssl/certs
* TLSv1.3 (OUT), TLS handshake, Client hello (1):
* TLSv1.3 (IN), TLS handshake, Server hello (2):
* TLSv1.3 (IN), TLS handshake, Encrypted Extensions (8):
* TLSv1.3 (IN), TLS handshake, Certificate (11):
* TLSv1.3 (IN), TLS handshake, CERT verify (15):
* TLSv1.3 (IN), TLS handshake, Finished (20):
* TLSv1.3 (OUT), TLS change cipher, Change cipher spec (1):
* TLSv1.3 (OUT), TLS handshake, Finished (20):
* SSL connection using TLSv1.3 / TLS_AES_256_GCM_SHA384
* ALPN, server accepted to use h2
* Server certificate:
*  subject: C=DE; O=HomeMatic; OU=1D8997144C; CN=homematic-raspi
*  start date: May 25 07:08:42 2022 GMT
*  expire date: May 22 07:08:42 2032 GMT
*  issuer: C=DE; O=HomeMatic; OU=1D8997144C; CN=homematic-raspi
*  SSL certificate verify result: self signed certificate (18), continuing anyway.
* Using HTTP2, server supports multi-use
* Connection state changed (HTTP/2 confirmed)
* Copying HTTP/2 data in stream buffer to connection buffer after upgrade: len=0
* Using Stream ID: 1 (easy handle 0x55a82a0d40e0)
> GET / HTTP/2
> Host: 192.168.5.53
> user-agent: curl/7.68.0
> accept: */*
> 
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* TLSv1.3 (IN), TLS handshake, Newsession Ticket (4):
* old SSL session ID is stale, removing
* Connection state changed (MAX_CONCURRENT_STREAMS == 8)!
< HTTP/2 200 
< accept-ranges: bytes
< cache-control: private, no-cache, must-revalidate, no-transform, max-age=0
< content-type: text/html; charset=iso-8859-1
< content-length: 2623
< date: Sun, 05 Jun 2022 11:47:59 GMT
< x-frame-options: SAMEORIGIN
< x-content-type-options: nosniff
< x-xss-protection: 1; mode=block
< x-robots-tag: none
< x-download-options: noopen
< x-permitted-cross-domain-policies: none
< referrer-policy: no-referrer
< 
<!DOCTYPE html">
[...]
```

So perhaps you can spot something fishy here. Not that this is a new / last-minute bug in the upcoming lighttpd 1.4.65 release...

Just in case you want to test this live again in a runinng RaspberryMatic installation. Here is the current ova update archive based on this PR:

https://cloud.light-speed.de/s/dYk6nRdyZw8Kpoi

gstrauss:
> So perhaps you can spot something fishy here. Not that this is a new / last-minute bug in the upcoming lighttpd 1.4.65 release...
> * http2 error: Invalid HTTP header field was received: frame type: 1, stream: 1, name: [], value: [u=2]

Good catch.  I missed updating some pre-cached tables when adding support for the upcoming RFC for HTTP/2 Priority (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-priority-12).  I need to test this further and add some HTTP/2 tests, but a preliminary patch is here: https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/e9565b2bb0197e54310e1637813c915fcfe27944

jens-maus:
> > * http2 error: Invalid HTTP header field was received: frame type: 1, stream: 1, name: [], value: [u=2]
> 
> Good catch. I missed updating some pre-cached tables when adding support for the upcoming RFC for HTTP/2 Priority (https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-priority-12). I need to test this further and add some HTTP/2 tests, but a preliminary patch is here.

Great that I could be of some here here for the lighttpd project. I just updated to your [e9565b2bb0197e54310e1637813c915fcfe27944](https://git.lighttpd.net/lighttpd/lighttpd1.4/commit/e9565b2bb0197e54310e1637813c915fcfe27944)  and could verify that HTTPS now works as expected with 1.4.65. So it seems your preliminary patch works as expected. Hope to see the official 1.4.65 released soon - what is the official schedule for this release, btw?

So I guess, regarding this PR we are finally done now, right? At least I can not remember of any further point to address. Great that you invested yourself so much time into giving us a help here in our nice little niche project. This is highly appreciated!

gstrauss:
> Great that I could be of some help here for the lighttpd project.

I very much appreciate your feedback.  Thank you.  Learning how others use lighttpd often helps me to improve lighttpd.

> Hope to see the official 1.4.65 released soon - what is the official schedule for this release, btw?

As soon as I roll the release. :)  ...It will be this week.  I want to add some tests for the preliminary patch, but the release is otherwise ready to go.

gstrauss:
The lighttpd configuration in this PR should measurably reduce the memory usage of lighttpd on RaspberryMatic at least a few kb (due to sharing the openssl SSL_CTX), and should be snappier (due to the auth.cache setting).

jens-maus:
Thank you. Will keep that in mind. And hopefully pur development paths will cross each other one day. Relaxing Pentecost to you and your family!

PR #2179, opened 2023-01-21
Correct MetaData MIN+MAX on SV change

jp112sdl:
### Description

When changing the type of a system variable, the `MetaData()` of `MIN` and `MAX` are not getting cleared, if the system variable has not a min-to-max range.

This patch will always set the `MetaData(MIN)` and `MetaData(MAX)` to `null` first.

### Related Issue

https://homematic-forum.de/forum/viewtopic.php?f=65&t=77676&start=20#p754163

### Types of changes

### Contributing checklist

### DIFF
--- a/buildroot-external/patches/occu/0097-WebUI-SetSysvarValueByUser.patch
+++ b/buildroot-external/patches/occu/0097-WebUI-SetSysvarValueByUser.patch
@@ -4,8 +4,8 @@
    !sv.ValueSubType(ist);
    sv.DPInfo( system.GetVar("sInfo") );
    sv.ValueUnit( system.GetVar("sUnit") );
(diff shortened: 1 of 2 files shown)


jens-maus:
Changed your PR to better use `RemoveMetaData()` instead of using 'null' because using null will just set it to `null` and not remove it!

jp112sdl:
Didn't know the presence of `RemoveMetaData`...
Requesting `MIN` or `MAX` gave back `null`. So I thought it simply has to be set to `null`.

MichaelN0815:
Wir müssen da noch mehr "resetten":
Unbenannt
Das sind "Überbleibsel" wenn die SV vorher Typ Logik oder Typ Werteliste war.

Issue #1191, opened 2021-03-22
Not enough real-time shares (cpu.rt_runtime_us == 0) with Docker


eloo:
**Describe the bug**
Hi,

i'm currently trying the docker (oci) image to replace my debmatic installation in future.
But i'm having some trouble to get it running. I'm also following the home-assistant addon issue here in github which looks very promising.

When i start the docker image (using compose) my rf-mod gets detected but it looks like rfd has issues to get started.

I have also tested to start rfd from within the container but also this does not work. 

Docker logs:
```
Starting watchdog...
Identifying onboard hardware: oci, OK
Initializing RTC Clock: onboard, OK
Running sysctl: OK
Checking for Factory Reset: not required
Checking for Backup Restore: not required
ln: /dev/raw-uart: File exists
Initializing System: OK
Starting logging: OK
Populating /dev using udev: done
Init onboard LEDs: init, OK
Starting irqbalance: OK
Starting network: eth0: link up, fixed, firewall, inet up, 172.23.0.2, OK
Identifying Homematic RF-Hardware: ....HmRF: RPI-RF-MOD/GPIO, HmIP: RPI-RF-MOD/GPIO, OK
Updating Homematic RF-Hardware: RPI-RF-MOD: 4.2.14, OK
Starting hs485dLoader: disabled
Starting xinetd: OK
Starting eq3configd: OK
Starting lighttpd: OK
Starting ser2net: disabled
Starting ssdpd: OK
Starting NUT services: disabled
Initializing Third-Party Addons: OK
Starting LGWFirmwareUpdate: ...OK
Setting LAN Gateway keys: OK
Starting hs485d: disabled
Starting multimacd: ERROR: not enough real-time shares (cpu.rt_runtime_us == 0)
Starting rfd: ....................ERROR
Starting HMIPServer: .......................................................................................................................................................ERROR
Starting ReGaHss: .OK
Starting CloudMatic: OK
Starting NeoServer: disabled
Starting Third-Party Addons: OK
Starting crond: OK
Setup onboard LEDs: booted, OK
```

System logs for rfd
```
Mar 22 23:45:11 homematic-raspi user.err monit[811]: 'rfd' process is not running
Mar 22 23:45:11 homematic-raspi user.info monit[811]: 'rfd' trying to restart
Mar 22 23:45:11 homematic-raspi user.info monit[811]: 'rfd' restart: '/etc/init.d/S61rfd restart'
[...]

PR #2386, opened 2023-08-11
Update real-time capability check for multimacd

honsma235:
### Description

Use more generic way to test if we can change real-time scheduling attributes.

On a kernel compiled with `CONFIG_RT_GROUP_SCHED=y` each cgroup domain which wants to use real-time scheduling would need to allocate rt_runtime_us.

If you are stuck with such a kernel one can disable the limit altogether changing the kernel parameter with `sudo sysctl -w kernel.sched_rt_runtime_us=-1` or
`echo "kernel.sched_rt_runtime_us = -1" | sudo tee /etc/sysctl.d/10-disable-rt-group-limit.conf` to make it permanent.

The previous check for real-time capability fails in this case, because it only checks for rt_runtime_us > 0 in the cgroup settings which don't reflect the kernel parameter.

### Related Issue
home-assistant/operating-system#1235
home-assistant/operating-system#2168
#1191

### Types of changes

### Alternate Designs
None considered

### Possible Drawbacks
Not that I'm aware of

### Verification Process
Implemented and tested locally on a Radxa 4SE with Debian Bullseye and Home Assistant supervised install

### Release Notes
N/A

### Contributing checklist

### DIFF
--- a/buildroot-external/overlay/RFD/etc/init.d/S60multimacd
+++ b/buildroot-external/overlay/RFD/etc/init.d/S60multimacd
@@ -107,7 +107,7 @@ start() {
         echo "ERROR: /dev/eq3loop missing"
       fi
     else
-      echo "ERROR: not possible to change real-time scheduling attributes"
+      echo "ERROR: real-time scheduling failed. Real-time group scheduling CONFIG_RT_GROUP_SCHED enabled, but no run time assigned?"
     fi
   else
     echo "not required"
--- a/buildroot-external/overlay/RFD/etc/init.d/S60multimacd
+++ b/buildroot-external/overlay/RFD/etc/init.d/S60multimacd
@@ -82,12 +82,10 @@ start() {
   if [[ -n "${HM_HMIP_DEV}" ]] && [[ -n "${HM_HMRF_DEV}" ]] &&
      ( ( ! echo "${HM_HMIP_DEV}" | grep -q "HMIP-RFUSB" ) || [[ "${HM_HMRF_DEV}" != "HM-CFG-USB-2" ]] ); then
 
-    # we need to check if the system has CONFIG_RT_GROUP_SCHED enabled
-    # and enough cpu.rt_runtime_us shares assigned or otherwise
+    # we need to check if the system allows changing real-time scheduling attributes or otherwise
     # multimacd startup will fail as it uses sched_setscheduler() to
     # raise the real-time priority for processing things with low latency
-    if [[ ! -e /sys/fs/cgroup/cpu/cpu.rt_runtime_us ]] ||
-       [[ $(cat /sys/fs/cgroup/cpu/cpu.rt_runtime_us) -gt 0 ]]; then
+    if chrt -f 10 echo >/dev/null 2>&1; then
 
       # run init function
       init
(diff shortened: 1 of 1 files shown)


jens-maus:
Thx for your contribution. I do have two small comments/requests before accepting this PR:

1. Please check and elaborate that the `chrt` commands are available for all supported RaspberryMatic platforms.
2. You changed the final ERROR output and do not provide some additional technical hint in how exactly a user could fix that issue or how the error could be interpreted correctly or actually googled. So please rephrase that final ERROR in case the cgroups feature/setting is missing so that user get some more technical hints what could actually be the reason or potential solution.

jens-maus (changes requested):
Please adapt as requested

honsma235:
As for the availability of chrt:
It is part of Busybox and enabled in Busybox.config https://github.com/jens-maus/RaspberryMatic/blob/c6a4faad9d3cf21ce4083258b642ef0f3025be3a/buildroot-external/Busybox.config#L611 so I guess it should be available in all builds.

But not sure where the file comes from. The header says something about automatically generated.

buildroot-external/overlay/RFD/etc/init.d/S60multimacd
```diff
@@ -109,7 +107,7 @@ start() {
         echo "ERROR: /dev/eq3loop missing"
       fi
     else
-      echo "ERROR: not enough real-time shares (cpu.rt_runtime_us == 0)"
+      echo "ERROR: real-time scheduling failed. Real-time group scheduling CONFIG_RT_GROUP_SCHED enabled, but no run time assigned?"
```
jens-maus, line 110: Please change to:

`ERROR: real-time scheduling test failed. CONFIG_RT_GROUP_SCHED not set or host kernel unable to utilize sched_setscheduler()`

jens-maus (changes requested):

buildroot-external/overlay/RFD/etc/init.d/S60multimacd
```diff
@@ -82,12 +82,10 @@ start() {
...
      ( ( ! echo "${HM_HMIP_DEV}" | grep -q "HMIP-RFUSB" ) || [[ "${HM_HMRF_DEV}" != "HM-CFG-USB-2" ]] ); then
 
-    # we need to check if the system has CONFIG_RT_GROUP_SCHED enabled
-    # and enough cpu.rt_runtime_us shares assigned or otherwise
+    # we need to check if the system allows changing real-time scheduling attributes or otherwise
     # multimacd startup will fail as it uses sched_setscheduler() to
     # raise the real-time priority for processing things with low latency
-    if [[ ! -e /sys/fs/cgroup/cpu/cpu.rt_runtime_us ]] ||
-       [[ $(cat /sys/fs/cgroup/cpu/cpu.rt_runtime_us) -gt 0 ]]; then
+    if chrt -f 10 echo >/dev/null 2>&1; then
```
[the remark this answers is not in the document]
honsma235: I don't have access to my setup for the upcoming days to verify the path. So I trust you on this:)

buildroot-external/overlay/RFD/etc/init.d/S60multimacd
honsma235: Did you mean "CONFIG_RT_GROUP_SCHED set" or "CONFIG_RT_GROUP_SCHED enabled"?
Otherwise it doesn't make sense to me.

PR #2399, opened 2023-08-22
fix missing tailscale dns, because of RO filesystem

konne:
### Description

Because of the RO filesystem the normal mechanism for tailscale magicDNS does not work.

You can see it for example if you run the command
`tailscale status`

it shows some lines and than the error:

```bash
#     - dns-os: writing to "/etc/resolv.pre-tailscale-backup.conf" in rename of "/etc/resolv.conf": open /etc/resolv.pre-tailscale-backup.conf: read-only file system
#     - dns: writing to "/etc/resolv.pre-tailscale-backup.conf" in rename of "/etc/resolv.conf": open /etc/resolv.pre-tailscale-backup.conf: read-only file system

```

### Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)
- [X] New feature (non-breaking change which adds functionality)

### Release Notes

add tailscale magicDNS support with the readonly filesystem

### Contributing checklist

- [X] My code follows the code style of this project.

### DIFF
--- a/buildroot-external/package/tailscale/S46tailscaled
+++ b/buildroot-external/package/tailscale/S46tailscaled
@@ -56,25 +56,27 @@ case "${1}" in
 		#     - dns-os: writing to "/etc/resolv.pre-tailscale-backup.conf" in rename of "/etc/resolv.conf": open /etc/resolv.pre-tailscale-backup.conf: read-only file system
 		#     - dns: writing to "/etc/resolv.pre-tailscale-backup.conf" in rename of "/etc/resolv.conf": open /etc/resolv.pre-tailscale-backup.conf: read-only file system
 
-		# fetch the current tailscale domain
-		domain=$(nslookup $(tailscale ip -1) 100.100.100.100 | grep name | sed -e 's/.*name = [^.]*\.//' -e 's/\.$//')
-		echo "add search domain: $domain"
+		# Append the nameserver directive to resolv.conf
+		echo "nameserver 100.100.100.100" | tee -a /var/etc/resolv.conf > /dev/null
 
 		# Extract the current search domains from resolv.conf
-		current_search=$(grep '^search ' /var/etc/resolv.conf | sed 's/^search //')
+		current_search=$(grep '^search ' /var/etc/resolv.conf | /bin/sed 's/^search //')
+
+		# fetch the current tailscale domain
+		sleep 1
+		domain=$(nslookup $(tailscale ip -1) 100.100.100.100 | /bin/grep name | /bin/sed -e 's/.*name = [^.]*\.//' -e 's/\.$//')
+		echo "add search domain: $domain"
 
 		# Combine the new domain with the existing search domains
 		new_search="$current_search $domain"
 
 		# Update if it already have a search line or add the search line
 		if [[ -n "$current_search" ]]; then
-			sed -i "s/^search .*/search $new_search/" /etc/var/resolv.conf
+			sed -i "s/^search .*/search $new_search/" /var/etc/resolv.conf
 		else
-			echo "search $new_search" | tee -a /etc/var/resolv.conf > /dev/null
+			echo "search $new_search" | tee -a /var/etc/resolv.conf > /dev/null
 		fi
 
-		# Append the nameserver directive to resolv.conf
-		echo "nameserver 100.100.100.100" | tee -a /etc/var/resolv.conf > /dev/null
 	;;
 	stop)
 		echo -n "Stopping ${DAEMON}: "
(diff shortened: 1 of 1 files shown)


konne:
@jens-maus currently a draft, I need to further check it, but you can already comment on.

@jens-maus a side question I have an even more complex topic around automatic tailscale certs. What you think is the best why for you to discuss the potential way. (github issues, github discussion, direct chat, ... something else)

konne (commented):
fix lint errors

buildroot-external/package/tailscale/S46tailscaled
```diff
@@ -50,12 +50,48 @@ case "${1}" in
...
 		# shellcheck disable=SC2086
 		${TAILSCALE_EXEC} up ${TAILSCALE_UP_ARGS} >/dev/null 2>&1 &
+
+		## adjust the resolv.conf manually because the original way from tailscale return an error see next comment lines
+		# Health check:
+		#     - dns-os: writing to "/etc/resolv.pre-tailscale-backup.conf" in rename of "/etc/resolv.conf": open /etc/resolv.pre-tailscale-backup.conf: read-only file system
+		#     - dns: writing to "/etc/resolv.pre-tailscale-backup.conf" in rename of "/etc/resolv.conf": open /etc/resolv.pre-tailscale-backup.conf: read-only file system
+
+		# fetch the current tailscale domain
+		domain=$(nslookup $(tailscale ip -1) 100.100.100.100 | grep name | sed -e 's/.*name = [^.]*\.//' -e 's/\.$//')
```
konne, line 60: @jens-maus can you help me.
If I run the line just on the ccu via SSH ... nslookup $(.....) ..... it works.
If this is running in the script I get the error "grep not found" / "sed not found"

konne:
@jens-maus is there any chance to get the lint errors in the message, so that I could just see it in the details

buildroot-external/package/tailscale/S46tailscaled:0: run 'shellcheck' and fix the warnings

jens-maus:
> @jens-maus currently a draft, I need to further check it, but you can already comment on.

ok, I converted this in a draft accordingly.

> @jens-maus a side question I have an even more complex topic around automatic tailscale certs. What you think is the best why for you to discuss the potential way. (github issues, github discussion, direct chat, ... something else)

If you have something working or with code then create a PR and we can discuss there. Otherweise use the github discussion fora in which there is a "Development" sub-category for these type of topics.

> @jens-maus is there any chance to get the lint errors in the message, so that I could just see it in the details

I already fixed the shellcheck error for you. So have a look if it works now.

However, regarding the general purpose of this PR: If `tailscale dns` / magicdns does not work because `/etc/resolv.conf` is the wrong location, then I would propose to potentially consider generating an upstream PR against tailscale to allow to define a different path for resolv.conf modifications. This seems to be a more reliable approach instead of trying to replicate what magicdns is doing.

konne:
@jens-maus thanks for your feedback. I already create a issue in the tailscale, but I don't expect a change. They already had a discussion longer ago around readonly resolv.conf and they solved it in a different way.
We could for now just fully disable at least the set, so that the error message disappear.  I would still like to have dynamically set the nameserver, that is not complex. I could not make the full set of the search stable and it was also like you have seen much to complex

jens-maus:
I adapted your PR draft again by having simplified your change a bit. Should work IMHO. Please test however. In addition, please also check if adding `--accept-dns=false` to the `TAILSCALE_UP_ARGS` variable still provides the necessary magicDNS functionality. Because I tend to add that additional up argument so that the up command does not try to modify `/etc/resolv.conf` itself and ends up in an error...

jens-maus:
Furthermore, please also test if simply symlinking `/etc/resolv.pre-tailscale-backup.conf` to `/var/etc/resolv.pre-tailscale-backup.conf` solves the initial error with tailscale not being able to modify `/etc/resolv.conf` contained in a read-only `/etc` path. Perhaps the issue can also be worked around by simply symlinking `/etc/resolv.pre-tailscale-backup.conf` to a place where the tailscale up command can then generated that tmp file.

konne:
@jens-maus I tried it, please double check, If I maybe did something wrong.

image

konne:
@jens-maus I tried you link way and it looks good. So I also linked the 

image

konne:
@jens-maus can you please help. I think it would be needed to be added to the tailscale.mk file, but I have no idea in which hook,...
```
		# relink resolv.pre-tailscale-backup.conf to /var/etc
                touch ../var/etc/resolv.pre-tailscale-backup.conf
		ln -snf ../var/etc/resolv.pre-tailscale-backup.conf $(TARGET_DIR)/etc/
```

PR #2424, opened 2023-09-04
feat: Changeable ID by user

indiana11011100:
- Next free ID or user choosen
- Numeric ID check with pve minimum 100

### Description

Changeable PVE-ID by user:
- Next free ID or user choosen
- Numeric ID check with pve minimum 100

### Types of changes

### Alternate Designs

No design change.

### Possible Drawbacks

ID Number check could fail on unexpected values (Should not happen with the RegEx Check).
Tested input numbers, numbers with letters, letters only, numbers below 100 and numbers already existing.
Please feel free to test again.

### Verification Process

Inspiration by "whiskerz007" is already given in the comments.

### Release Notes

- Changeable PVE-ID by user
- More flexibility for user structure

### Contributing checklist

### DIFF
--- a/scripts/install-proxmox.sh
+++ b/scripts/install-proxmox.sh
@@ -24,8 +24,10 @@ trap die ERR
 trap cleanup EXIT
 
 # Set default variables
-VERSION="1.11"
+VERSION="1.12"
 LOGFILE="/tmp/install-proxmox.log"
+PVE_NUMERIC_RE="^[1-9][0-9]+$"
+NEXTID=$(pvesh get /cluster/nextid)
 LINE=
 
 function error_exit() {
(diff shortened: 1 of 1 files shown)


jens-maus:
Please explain why you think this change is required! Don't really get it, because it is IMHO irrelevant which ID a certain VM will get. The only important point is that the `install-proxmox.sh` script uses the next free id (which it does) which makes sure there will be no overlap and actually outputs the chosen ID, which it also does. So IMHO this change/feature is not necessary. So please elaborate.

indiana11011100:
Moin Jens,
in  my case I like to sort my LXCs / VMs with some space between the numbers and between the different types. That gives me the ability to restore some backup or do some experimental LXC / VM right next to the original one. Also sorting LXC / VM by content type (e.g. Container for Network stuff, Apps, Coding, Labs, ...).

If the ID can't be choosen while setup, and PROXMOX offers this for any LXC / VM, the only way is to create, backup, destroy and restore with needed ID.

Hope you see the point and effort this could save.
BR indy

PR #2852, opened 2024-09-17
modified patch-ha-addon-macvlan.sh script to print default addon name  in interactive mode

clemenstyp:
### Description

I modified the patch-ha-addon-macvlan.sh script, so that in interactive usage the name of the addon container is printed into the console. To accept the default value the user just have to press enter. 

### Types of changes

### Release Notes

N/A

### Contributing checklist

### DIFF
--- a/scripts/patch-ha-addon-macvlan.sh
+++ b/scripts/patch-ha-addon-macvlan.sh
@@ -126,7 +126,7 @@ fi
 
 echo -n "RaspberryMatic Add-on Hostname (e.g. 5422eb72-raspberrymatic): "
 if [[ -z "${CCU_CONTAINER_NAME}" ]]; then
-  CCU_CONTAINER_NAME=$(docker ps --format '{{.Names}}' | grep "_raspberrymatic" | cut -c7-)
+  CCU_CONTAINER_NAME=$(docker ps --format '{{.Names}}' | grep "_raspberrymatic" | cut -c7- | awk '{print $0}')
   if [[ -z "${NON_INTERACTIVE}" ]]; then
     read -r -e -i "${CCU_CONTAINER_NAME}" CCU_CONTAINER_NAME </dev/tty
   else


scripts/patch-ha-addon-macvlan.sh
```diff
@@ -126,7 +126,7 @@ fi
 
 echo -n "RaspberryMatic Add-on Hostname (e.g. 5422eb72-raspberrymatic): "
 if [[ -z "${CCU_CONTAINER_NAME}" ]]; then
-  CCU_CONTAINER_NAME=$(docker ps --format '{{.Names}}' | grep "_raspberrymatic" | cut -c7-)
+  CCU_CONTAINER_NAME=$(docker ps --format '{{.Names}}' | grep "_raspberrymatic" | cut -c7- | awk '{print $0}')
```
jens-maus, line 129: Sorry, but I think this change does not make sense. `CCU_CONTAINER_NAME` should already contain the full docker container name after the `cut -c7-` . Or what does the additional `awk '{ print $0}'` try to solve? Please elaborate.

clemenstyp:
Yes, CCU_CONTAINER_NAME already contains the container name, but with awk '{print $0}' it gets printed into the console, so that the user just needs to press enter to accept it. So this ist only needed in the interactive version of the script   --Von unterwegs gesendetAm 23.09.2024 um 20:25 schrieb Jens Maus ***@***.***>:﻿
@jens-maus commented on this pull request.



In scripts/patch-ha-addon-macvlan.sh:
> @@ -126,7 +126,7 @@ fi
 
 echo -n "RaspberryMatic Add-on Hostname (e.g. 5422eb72-raspberrymatic): "
 if [[ -z "${CCU_CONTAINER_NAME}" ]]; then
-  CCU_CONTAINER_NAME=$(docker ps --format '{{.Names}}' | grep "_raspberrymatic" | cut -c7-)
+  CCU_CONTAINER_NAME=$(docker ps --format '{{.Names}}' | grep "_raspberrymatic" | cut -c7- | awk '{print $0}')

Sorry, but I think this change does not make sense. CCU_CONTAINER_NAME should already contain the full docker container name after the cut -c7- . Or what does the additional awk '{ print $0}' try to solve? Please elaborate.

jens-maus:
> Yes, CCU_CONTAINER_NAME already contains the container name, but with awk '{print $0}' it gets printed into the console, so that the user just needs to press enter to accept it. So this ist only needed in the interactive version of the script 

This rational is clear, but I doubt that this is really necessary since the actual printout of the variable is handled by the `read` command a few lines later (cf. https://github.com/jens-maus/RaspberryMatic/pull/2852/files#diff-edf8e5d7aa2073c72affc4c176b3c85f7c47ca315d72173af90c90531adad4a7R131). So I need to check that myself if your observation is correct since I am almost certain that this should already work like you expect it to work.

jens-maus:
Sorry, but after testing this myself, I don't see any issue with regard of automatically retrieving and outputting the currently running container name. See here:

Sep-24-2024 21-06-39

So if there is really an issue, please reopen this PR and demonstrate the issue with a similar screencast like above.

PR #3785, opened 2026-04-17
rework psplash startup to be able to restart psplash

jens-maus:
This adds psplash helper functions to rcS/rcK init files which ensure that psplash will run as soon as psplash write commands are executed. This should fix issues where psplash might not be able to start early enough due to framebuffer switch-over when DRM comes up. Now each write command will ensure that psplash runs.

## Summary by CodeRabbit

* **Chores**
  * Improved boot splash reliability and robustness (better handling of stale IPC and conditional shutdown)
  * Consolidated splash messaging to avoid noisy errors during startup/shutdown
  * Added boot sequence logging to capture init output for diagnostics
  * Refined initialization and shutdown ordering for more predictable startup/shutdown behavior

### DIFF
--- a/buildroot-external/overlay/base/etc/init.d/rcK
+++ b/buildroot-external/overlay/base/etc/init.d/rcK
@@ -4,8 +4,32 @@
 # Stop all init scripts in /etc/init.d
 # executing them in reversed numerical order.
 
-# start psplash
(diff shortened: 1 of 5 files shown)


jens-maus:
@coderabbitai resume

jens-maus:
@coderabbitai review

PR #3845, opened 2026-05-10
Add missing kernel config for netfilter to fix iptables legacy NAT

fl0x53:
Add missing options to kernel build config to fix missing NAT table in iptables legacy.

Current build of OpenCCU 3.87.6.20260509 introduced kernel builds > 6.17 that came with changes to iptables-legacy.
PR #3769 tried to fix that  #3765, but still results in:

```shell
#     - running [/usr/sbin/iptables -t nat -N ts-postrouting --wait]: exit status 3: iptables v1.8.11 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?)
Perhaps iptables or your kernel needs to be upgraded.
```
I added the most relevant build configs from [RaspiOS](https://raw.githubusercontent.com/raspberrypi/linux/refs/heads/rpi-6.18.y/arch/arm64/configs/defconfig) and checked against a current Fedora 44 config.

So let's give it a try 🤞

## Summary by CodeRabbit

* **Chores**
  * Core networking enabled: TCP/IP stack, IPv4/IPv6 support, IPv6 route info, bridge netfilter, and MPTCP with IPv6.
  * Connection tracking re-enabled with event notifications and IPv4 defragmentation.
  * Expanded NAT capabilities including redirect and masquerade support.
  * Broadened Netfilter/xtables and legacy IPv4/IPv6 iptables/arptables match and target modules to extend filtering and NAT.

[Review Change Stack](https://app.coderabbit.ai/change-stack/OpenCCU/OpenCCU/pull/3845)


jens-maus:
> ```shell
> #     - running [/usr/sbin/iptables -t nat -N ts-postrouting --wait]: exit status 3: iptables v1.8.11 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?)
> Perhaps iptables or your kernel needs to be upgraded.
> ```

Thanks for the PR, but where is this `iptables -t nat -N ts-postrouting` execution coming from? Because I can't see any reference to this iptables execution and also wonder partly why you need to setup a NAT firewall route anyway since there is hardly any use case for NAT networking in a CCU, isn't it?

fl0x53:
> > ```shell
> > #     - running [/usr/sbin/iptables -t nat -N ts-postrouting --wait]: exit status 3: iptables v1.8.11 (legacy): can't initialize iptables table `nat': Table does not exist (do you need to insmod?)
> > Perhaps iptables or your kernel needs to be upgraded.
> > ```
> 
> Thanks for the PR, but where is this `iptables -t nat -N ts-postrouting` execution coming from? Because I can't see any reference to this iptables execution and also wonder partly why you need to setup a NAT firewall route anyway since there is hardly any use case for NAT networking in a CCU, isn't it?

True, that's coming from tailscale (see `tailscale status`).
I'm actually having an issue with that missing NAT as I'm using the OpenCCU on a Raspi4 as an exit node to my home net. Home Assistant (cloud instance) is not able to control my Sonos speakers if NAT is not working properly. That worked fine until the last release with that kernel changes concerning iptables legacy.

jens-maus:
Ok, thanks for the clarification. Then I will check that and try to correct it for the next release.

fl0x53:
Thanks for removing those unnecessary extra options!  👍

PR #3950, opened 2026-06-14
security: restrict WebUI/services to local networks by default and harden port-forwarding detection

jens-maus:
This addresses the problem that OpenCCU centrals remain freely reachable (even without a password) from the internet despite the existing checkPortForwarding.sh watchdog.

Prevention (libfirewall.tcl, base + oci variant):
- Restrict otherwise publicly opened ports (WebUI 80/443, SSH, user defined ports and services with 'full' access) to local/private source networks by default in BOTH firewall modes (RESTRICTIVE and MOST_OPEN). External access (e.g. via a router port forwarding or a globally routed IPv6 address) is now dropped at the device, independent of the router configuration.
- The set of "local" source networks combines the well-known private/loopback/link-local ranges (RFC1918, CGNAT, IPv6 ULA/link-local) with the networks currently configured on the device's interfaces (incl. globally routed IPv6 prefixes used on the LAN and VPN tunnels), so any LAN setup keeps working and the user can never lock himself out.
- New override file /etc/config/AllowExternalAccess completely disables this local-only restriction for users who knowingly want direct external access. This can later be surfaced as a WebUI option (with a warning).

Detection (checkPortForwarding.sh):
- Fix: the previous external scan piped nmap into "awk -f <(...)", but busybox ash does not support process substitution, so OPEN_PORTS was always empty and the scan never detected anything. The awk program is now passed via a plain pipe.
- Add a reliable local configuration audit that detects the most critical case (WebUI reachable from the internet without a password = external access allowed AND authentication disabled) without depending on an unreliable self-scan.
- Add a firewall protection state audit that verifies the OPENCCU_LOCAL gate chain is actually active when it is supposed to be, so a silently failed prevention (e.g. firewall not applied) is detected instead of leaving the device unprotected without anybody noticing.
- Make the CCU fingerprint match whitespace tolerant, use nmap -T4 (fewer missed ports than -T5) and log explicitly when the public IP cannot be determined or nmap is missing (fail-safe instead of a false all-clear).
- Run the check daily instead of weekly for faster detection.

## Summary by CodeRabbit

* **New Features**
  * Added an “Advanced settings” button to the WebUI control panel with an Advanced Settings dialog.
  * Added an “Allow external internet access” option to control whether remote WebUI/API access is permitted.
  * Enabled a local-only access gate by default to restrict externally opened ports (SSH/WebUI and full-access services) to local networks.
  * Applied the local-only firewall behavior to IPv6 when available.
* **Bug Fixes**
  * Strengthened port-forwarding exposure detection with additional security audits and improved detection behavior.


Baxxy13:
> * Run the check daily instead of weekly for faster detection.

That one seems to be missing...

https://github.com/OpenCCU/OpenCCU/blob/723d7ece54b3348f0afc341b88249c27aac7731e/buildroot-external/overlay/base-openccu/etc/crontab.root#L9

jens-maus:
Yes, I reverted it back to weekly as I still feel this should be enough. Isn't it?

PR #3967, opened 2026-06-21
WebUI: escape systemName before ReGaScript interpolation in cp_advancedsettings.cgi (root RCE fix)

claymore666:
Fixes an authenticated ReGaScript injection → root RCE in the WebUI ControlPanel **Advanced Settings** page.

`cp_advancedsettings.cgi`'s `set_systemname` built a ReGaScript string by interpolating the user-supplied `systemName` directly into a single-quoted string literal:

```tcl
set isecmd "system.Name('$systemname');"
```

and executed it via `rega_script`, i.e. in ReGaHss running as **root**. A `systemName` containing a single quote terminated the literal and the remainder was parsed as ReGaScript, allowing an authenticated admin to inject arbitrary ReGaScript (e.g. `system.Exec(...)`) and execute commands as root. `action_save_settings` calls `set_systemname` before any other processing, so the injection fired on every `save_settings` request regardless of the other form fields. On a default install the Admin WebUI account has an empty password, making the precondition trivial to meet.

**Fix:** escape the two ReGaScript string-literal metacharacters before interpolation, in a single `string map` pass (`\` → `\\`, `'` → `\'`):

```tcl
set systemname [string map {\\ \\\\ ' \\'} $systemname]
```

Legitimate names containing apostrophes round-trip unchanged; the injection payload is stored verbatim as the literal system name. Verified end-to-end against a live instance: the PoC payload no longer executes and no canary file is created.

The patch is applied to both copies of the file under the `0135-WebUI-Add-ControlPanel-AdvancedSettings` patch directory.

Closes #3966
Refs: GHSA-6qrx-4wv4-4534

## Release Notes

* **New Features**
  * Added "Advanced Settings" button to Control Panel for configuring system options including watchdog checks, system name, backup settings, feature toggles, and storage checks.
  * Browser page titles now dynamically display your system name or default to "OpenCCU WebUI" when unnamed.
  * Extended tooltip and help system support for advanced settings.

Issue #3819, opened 2026-04-28
Settings>Favorites "add program" fails with single tick in value field of first THEN statement


Joerg-rw:
### Describe the issue you are experiencing

In "Home page > Settings > Favourites > Create favourites" the "Add program" selection list of programs comes up empty.
Cause: a single-tick ( ' ) in value field of first statement in THEN which breaks the  dynamic assembly of the "Creating favourite - program selection" list

<img width="1900" height="338" alt="Image" src="https://github.com/user-attachments/assets/9f686b16-e3e0-4ed7-a689-88a48df66442" />

### Describe the behavior you expected

no illegal chars in value field of assignemnt statements, neither in conditions and ideally neither in description text nor in program name. Particularly ' (tick) obviously needs escaping or other special handling.

Noteworthy: in "Home page > Programs and connections > Programs" there's obviously no such problem

### Steps to reproduce the issue

1. create a program like this:
<img width="1900" height="338" alt="Image" src="https://github.com/user-attachments/assets/f6cffa8f-481c-47a3-9b49-8a1dad672a01" />
<img width="1900" height="423" alt="Image" src="https://github.com/user-attachments/assets/4033b0e8-f30e-40ce-a85b-f98926aad93c" />

2. proceed to  "Home page > Settings > Favourites > Create favourites"
3. click "Add program" button
4.  list comes up empty: 
<img width="1900" height="780" alt="Image" src="https://github.com/user-attachments/assets/36c7df7a-7469-421d-9cb1-15865565e647" />

### What is the version this bug report is based on?

any Raspimatic / OpenCCU

### Which base platform are you running?

rpi3 (RaspberryPi3, ARM64/aarch64)

### Which HomeMatic/homematicIP radio module are you using?

n/a

### Anything in the logs that might be useful for us?

```txt
no
```

### Additional information

related thread: https://homematic-forum.de/forum/viewtopic.php?f=65&t=87900&p=858614#p858614

claymore666:
The ticks currently fail because they are not properly encoded, see #3972. Workaround is putting a \ before each one like so: `\'test\'`
As fixing this involves the script engine which should be tested properly  we havent proposed a fix yet.

PR #3972, opened 2026-06-21
WebUI: fix ReGaScript injection root RCE (multiple sinks)

### LINKED ISSUE #3971: Authenticated ReGaScript injection → root RCE in WebUI (multiple sinks)
## Summary

The OpenCCU WebUI builds HomeMatic Script (ReGa) by interpolating user-supplied strings into `var name = "<value>";` declarations. Insufficient — or, in the legacy PDA interface, entirely absent — escaping lets an authenticated **LEVEL USER** break out of the generated string literal so the remainder is parsed as live HomeMatic Script, reaching `system.Exec` and running OS commands **as root** (ReGaHss runs as `uid=0`).

This is the same class as GHSA-6qrx-4wv4-4534 (the `cp_advancedsettings.cgi` sink, already fixed in #3967), and the upstream report eq-3/occu#131 — but here it's **a pattern, not a single function**, with three instances. Originally reported privately as **GHSA-c45w-fgwq-mvq4**; filing publicly at the maintainer's request so the fix can be merged via a normal PR. Sanitized writeup — turnkey payloads omitted.

## Affected

OpenCCU WebUI as shipped (`occu`, pinned in `buildroot-external/package/occu/occu.mk`; `OCCU_VERSION = 3.89.2-1` at time of writing). No patch under `buildroot-external/patches/occu/` addresses any of the sinks below.

## Findings

**1 — `hmscript_escapeString`, incomplete escaping (central JSON-RPC path).**
`WebUI/www/api/eq3/hmscript.tcl` escapes `'`, `"` and control chars but **not the backslash**. `hmscript` emits `var <name> = "[hmscript_escapeString $value]";`. Because ReGa does its own backslash decoding, a crafted value yields a literal backslash followed by an unescaped quote, terminating the literal early. The JSON-RPC API funnels user input through this one helper, so it covers the whole JSON-RPC surface.

**2 — `rega_escape`, identical latent bug.**
`WebUI/www/tcl/eq3/rega.tcl` → `proc rega_escape` has the same missing-backslash escaping. It has **no callers** in the WebUI tree today, so it's not presently reachable — but it's the same broken template waiting for a future caller.

**3 — PDA interface (`www/pda/**`), no escaping at all.**
The legacy PDA UI builds `var x = "$id";` from imported CGI params with **no escaper** (e.g. `www/pda/eq3/favorite.tcl` `Favorite_get`, reached via `www/pda/fav.cgi`). `cgi_import` does no sanitization and there's no numeric/format validation on the object-id params, so a bare `"` is sufficient. The same unescaped pattern appears at ~36 sites under `www/pda/**`, reachable via `fav.cgi` / `favlist.cgi`. `/pda` is served as CGI by lighttpd.

## Preconditions

An authenticated WebUI session at **LEVEL USER** (privilege 2) — the JSON-RPC path and the PDA `session_isValid` gate both require only a valid session, not ADMIN. Note that on a default install the Admin account has an empty password.

## Impact

Arbitrary OS command execution as **root** from an authenticated (least-privilege `USER`) WebUI session. **Critical.**

## Fix direction

The two distinct mechanisms each need their own fix:

- **Findings 1 & 2 (escaper):** add backslash escaping as the **first** `string map` entry (`"\\" "\\\\"`) in both `hmscript_escapeString` and `rega_escape`, so the value can no longer break out of the literal.
- **Finding 3 (PDA):** escaping the 36 sinks is fragile; instead **whitelist-validate the object-id params at the CGI entry** — gate `favId`/`favListId` in `fav.cgi`/`favlist.cgi` with `if {![regexp {^[0-9]+$} $x]} { set x "" }`. The PDA `*.cgi` endpoints that import object-id params are `fav.cgi` and `favlist.cgi`; the others import no id params, so the sink sites are then unreachable with attacker-controlled non-numeric input.

A fix implementing exactly this exists in the private advisory fork and can be opened as a public PR against `master`.

Refs: GHSA-c45w-fgwq-mvq4, eq-3/occu#131

claymore666:
Closes #3971.

Public fix for **GHSA-c45w-fgwq-mvq4** (authenticated ReGaScript injection → root RCE in the WebUI, multiple sinks), filed at the maintainer's request — same treatment as #3967 for GHSA-6qrx. The change is ported from the private advisory fork (`OpenCCU/OpenCCU-ghsa-c45w-fgwq-mvq4`) and rebased onto current `master`.

### What it fixes
Two distinct mechanisms, each handled at the right layer:

- **Findings 1 & 2 (escaper, `hmscript_escapeString` + `rega_escape`):** add backslash escaping as the **first** `string map` entry (`"\\" "\\\\"`). Without it, ReGa's own backslash decoding let a crafted value emit a literal `\` before an unescaped `"`, terminating the generated `var x = "...";` literal early. `hmscript_escapeString` covers the whole JSON-RPC surface; `rega_escape` is the latent twin.
- **Finding 3 (PDA, `www/pda/**`):** the ~36 `var x = "$id";` sinks have no escaper, so they're fixed by **whitelist-validating the object-id params at the CGI entry** rather than escaping each sink — `fav.cgi` gates `favListId`/`favId`, `favlist.cgi` gates `favListId`, via `if {![regexp {^[0-9]+$} $x]} { set x "" }`. Those are the only PDA `*.cgi` endpoints that import object-id params, so the sinks become unreachable with attacker-controlled non-numeric input.

### Verification
- The `0207` patch applies cleanly to the currently-pinned occu (`OCCU_VERSION = 3.89.2-1`) — the patch's `.orig` baselines are byte-identical to the four target files at `OpenCCU/occu@3.89.2-1` (`hmscript.tcl`, `rega.tcl`, `fav.cgi`, `favlist.cgi`). CI's package-consistency (`make check`) will re-confirm.
- A second commit adds a self-contained regression test (`tests/security/rega_script_injection_test.tcl`, runs under `tclsh` with no firmware build) plus a `VERIFICATION.md` describing the security invariant. Drop that second commit if it's out of scope for this repo; the fix itself is the first commit and stands alone.

Refs: GHSA-c45w-fgwq-mvq4, eq-3/occu#131

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of special characters so backslashes are preserved safely in WebUI script data.
  * Tightened validation for favorite and list identifiers to block invalid input and reduce script-injection risk.
  * Fixed WebUI behavior for favorite-related actions, including program execution handling.

* **Documentation**
  * Added security verification notes and regression test coverage for the injection fix.


claymore666:
Addressed the CodeRabbit review in f429af5:
- **fav.cgi typo** — renamed `executePorgram` → `executeProgram` (the `COMMANDS(exec)` dispatch entry and the proc; it's the only reference, the dispatch is file-local) and regenerated `0207.patch` via `create_patches.sh` so the committed patch stays byte-consistent. Done byte-safely so the latin1 `fav.cgi` encoding is preserved.
- **VERIFICATION.md** — added code-fence languages (MD040) and reconciled the OCCU version note (`.orig` baselines are byte-identical to the pinned `3.89.2-1`; unchanged since `3.87.6-3`).

buildroot-external/patches/occu/0207-WebUI-Fix-ReGaScriptInjection.patch
```diff
@@ -0,0 +1,65 @@
+--- occu/WebUI/www/api/eq3/hmscript.tcl.orig
++++ occu/WebUI/www/api/eq3/hmscript.tcl
+@@ -59,6 +59,7 @@
+ 
+ proc hmscript_escapeString { str } {
+   return [string map {
++    "\\" "\\\\"
```
jens-maus, line 7: Even thought that this might fix some security flaw, I am curious if this won't affect the rega scripting functionality in the WebUI itself. So perhaps we need somewhat more testing here? @Baxxy13 could you potentially investigate a but and see if backslash uses (`\`) and escaping in the WebUI itself still works after adding this additional escaping to the `hmscript_escapString()` and `rega_escape()` functions?

buildroot-external/patches/occu/0207-WebUI-Fix-ReGaScriptInjection.patch
```diff
@@ -0,0 +1,65 @@
...
+ 	if { {true} != [session_isValid $sid] } then {
+ 		http_head
+ 		puts [eval [template_parseFile $TEMPLATE_FILE_ERROR_SESSION]]
+--- occu/WebUI/www/tcl/eq3/rega.tcl.orig
++++ occu/WebUI/www/tcl/eq3/rega.tcl
+@@ -13,6 +13,7 @@
+ ##
+ proc rega_escape {value} {
+   return [string map {
++    "\\"  "\\\\"
```
jens-maus, line 62: same here! I think this definitely needs more investigation before we can merge this... @Baxxy13 can you perhaps test that to some extend?

jens-maus (changes requested):

claymore666:
I went ahead and tested this on a dev instance (3.87.6.20260614): deployed the patched `hmscript.tcl` over the live file and exercised it through the real JSON-RPC API — creating system variables with backslash names via `SysVar.createBool` and reading the stored name back.

The extra escaping doesn't break backslash handling, it actually fixes it. The current escaper already mangles legitimate backslashes, because they reach ReGa unescaped and `\t`, `\n`, … get parsed as control chars:

- `a\test\b` → stored as `a<TAB>est\b` (the `\t` becomes a TAB)
- `ends\` → creation fails (the lone `\` escapes the closing quote)
- `x\";WriteLine(31337);!` → the injected `WriteLine(31337)` actually runs

With the patch all three are correct: `a\test\b` and `ends\` round-trip exactly, and the last one is stored verbatim as inert data (no execution). The reason it's safe: ReGa decodes `\\` → `\` when parsing `"..."`, so doubling the backslash in the escaper and halving it in the parser cancel out — legitimate `\` survives unchanged.

`rega_escape` gets the same one-character fix but has no callers today, so it's covered by the same reasoning rather than a live test. I added the before/after as a "Layer 4" section in `tests/security/VERIFICATION.md`.

jens-maus:
BTW: Bleiben wir erst einmal im Deutschen, solange hier nur Deutsche involviert sind ;-)

> I went ahead and tested this on a dev instance (3.87.6.20260614): deployed the patched `hmscript.tcl` over the live file and exercised it through the real JSON-RPC API — creating system variables with backslash names via `SysVar.createBool` and reading the stored name back.

Danke fürs weitere testen. Die JSON-RPC API ist aber nur eine möglichkeit HMScript/ReGa-Skripting zu senden. Schau dir mal an was im Repo so unter "scripts/testcases" (https://github.com/OpenCCU/OpenCCU/tree/master/scripts/testcases) liegt. Auch muss man nochmal testen (ich denke @Baxxy13 wird hier sicherlich auch noch einmal aktiv werden) direkt mit der WebUI mal schauen was da so passiert unter "Skript testen" und in der Eingabemaske der Programme wenn man HM/ReGa-Skripte dort verwendet. Das ganze ist leider ziemlich historisch gewachsen, ist ziemlich undurchsichtig und sehr fragil erfahrungsgemäß. Deshalb will ich hier ungern ein Schnellschuss machen und erst mal konkret auch solche Endusersachen wie via der WebUI (und nicht nur via externer APIs) das ganze sauber durchtesten. Auch gibt es meines Wissens (kannst du gerne durchforsten) den einen oder anderen offenen Bug/Issue-Report im Repo der sich auf ähnliche Escape-Probleme bezieht. Vielleicht repariert dein vorgeschlagener Fix ja das gleich mit.

claymore666:
Ja den Eindruck hatte ich auch daß andere, alte Issues teilweise gleiche Ursache haben könnten, ließ sich gestern aber nicht definitiv beweisen. Aber dann halten wir den PR doch erstmal zurück und ich erfasse erst nochmal für mich was alles nötig ist zu testen, und würde dann nochmal die Testdocu anpassen, Wenn ein erfahrener Contributor noch mittesten würde, wäre ich dankbar.

jens-maus:
> ... Wenn ein erfahrener Contributor noch mittesten würde, wäre ich dankbar.

So machen wir es. @Baxxy13 wird hier sicherlich irgendwann aktiv werden und erst vortests machen. Wenn die nichts zu tage fördern würde ich dafür plädiere den PR dann ggf. doch einfach zu mergen und die Nightly Build tester (sind inzwischen schon einige) den Rest erledigen zu lassen. Aktuell ist der 3.89.x release ja noch in etwas weiterer Ferne und da bleibt dann ggf. genug Zeit etwaige Regressions mit hilfe der nightly build tester aufzudecken.

claymore666:
Hi Jens,

hat mir jetzt keine Ruhe gelassen und ich habe mir nochmal die anderen Wege angeschaut.

Imho gibt es genau eine Stelle, an der die WebUI Nutzerdaten in ein Skript hineinbaut – `hmscript_escapeString` in `hmscript.tcl`. Über die läuft das `var name = "<wert>";` für alle ~80 JSON-RPC-Methoden (jeder Aufruf geht über `hmscript $script args`). Genau diese Stelle ändert der Patch. Alles andere (siehe unten) escapt gar nicht und wird vom Patch nicht angefasst.

**Meine Tests:**

**1. Der Escaper direkt gegen die echte ReGa-Engine** (mit `tclsh` + `tclrega.so`, im Stil deiner `tcl-test.tcl`). Das ist der Pfad, den alle API-Methoden teilen. Jeweils alt vs. neu:

| Eingabe | alt (ungepatcht) | neu (gepatcht) |
|---|---|---|
| `a\tb` | `a<TAB>b` – Daten kaputt | `a\tb` ✓ |
| `ends\` | Skript bricht ab | `ends\` ✓ |
| `\";WriteLine(31337);!` | **31337 läuft (RCE)** | harmlos ✓ |

**2. Echte WebUI-API, vorher/nachher** (gepatchte Datei per bind-mount eingehängt), über `SysVar.createBool/createFloat/createEnum`. Gleiches Bild: Backslash-Namen sind ungepatcht kaputt oder die Anlage scheitert – gepatcht stimmen sie exakt.

**3. Encoding (deine `scripts/testcases`):** Die zeigen fest auf Gerät 1645, das es auf meiner sauberen VM nicht gibt. Der Umlaut-Test hängt aber am Kanal, nicht am Objekt – also auf eine Sysvar `zenc_RFö` umgebogen und deine drei Kanäle mit hexdump geprüft. Überall korrekt ISO-kodiert (`ö` = `f6`, bzw. `RF%F6`). **Keine Encoding-Regression.**

**4. „Skript testen":** läuft über `ReGa.runScript` auf dem **rohen** Skript, ganz ohne Escaping (so gewollt). Vorher/nachher identisch – der Patch fasst das nicht an.

**Noch ein Fund – du hattest recht:** Es gibt einen verwandten Bug, #3819 (Hochkomma in einem THEN-Wert killt die Favoriten-Programmliste). Den hab ich nachgestellt. Ursache ist `programChooserArray.htm` – dort werden `Name()`, `PrgInfo()` und der THEN-Text **ohne Escaping** in einfach-gequotete JS-Strings geschrieben:

```js
tmp['prginfo'] = 'desc with ' single tick';   // String endet beim ' → JS kaputt → Liste leer
```

Das ist eine **eigene Stelle** (ReGaHss-ESP, nicht `hmscript_escapeString`). Der aktuelle Patch fixt das also **nicht** – das wäre ein separater Fix.

Die „Eingabemaske der Programme" selbst schreibt Namen/Werte über ReGaHss (C++), nicht über unseren Tcl-Escaper. Das interaktive Durchklicken überlass ich gern @Baxxy13.

jens-maus (approved):

PR #4125, opened 2026-08-20
add optional per-Home-Assistant-user session/credential storage

sandmaennchen5:
This pull request introduces optional per‑Home‑Assistant‑user session and credential storage for the OpenCCU Home Assistant add-ons and proxy components.

The goal is to improve user‑specific authentication handling by allowing each Home Assistant user to maintain their own session data instead of relying on shared or global credentials. This enhances security, supports multi‑user environments, and enables more granular control over access to CCU resources.

## Summary by CodeRabbit

* **New Features**
  * Added optional per-user persistence for OpenCCU WebUI sessions.
  * Added optional encrypted credential storage with automatic re-login after session expiration.
  * Added session keep-alive, logout, invalidation, and credential management improvements.

* **Configuration**
  * Added settings for remembered sessions, remembered credentials, and keep-alive intervals.
  * Added validation requiring keep-alive intervals between 1 and 599 seconds.
  * Added English descriptions covering session behavior and credential-storage security considerations.


sandmaennchen5:
@coderabbitai resume

jens-maus (changes requested):
Thanks @sandmaennchen5 for this PR. I quickly looked over it (not tested yet), but do have some first general comments/questions before I will do a deeper review of it:

AFAICS this PRs tries to add functionality that more or less no dedicated login might be required anymore because it saves the credentials (username+password) in an encrypted storage under `/usr/local/etc/config`, so in the localfs of the OpenCCU context. Right? While I do understand the principle idea behind that, I still have my doubts and concerns if this is really something utterly important. Why isn't the simple session id cookie storage not enough here? Your PR reasoning is that in a HA multi-user environment there might be an issue. But which issue exactly? If you really have different users that log into your HA via different credentials, then they would anyway use different browsers and thus would have different session IDs when opening their OpenCCU app/add-on. So AFAIK it cannot happen that user A uses the session ID of user B if they are really using different browsers. In addition, the way you are currently trying to get the username+password and storing it into an encrypted credential store within OpenCCU is somewhat IMHO a bit overcomplicated or lets say "introduced quite some amount of code to retrieve the username+password from a logging in procedure". In addition, I do not like the idea of storing these credentials in the /usr/local/etc/config localfs space as this (even thought it is encrypted) constructs some security concerns and IMHO it is also not required to store it there because it does not need to survive any reboot IMHO. So why it is really necessary that you are storing the full username+password and do not simply rely on the already implemented session ID storage via cookies? If only the multi-user concerns is affected here then why don't you simply change the session cookie storage to include the username and then have multiple cookies so that within the same browser it might choose a different session ID.

As said, I do like your further changes to ha-proxy.js, but I also do have my concerns here if they really help in the end, because to me they just look like you are trying to implement a way that no additional login to the OpenCCU instance is required once the HA addon/app is used. And I am not sure if this is really worth all the changes you are proposing. So please lets discuss a bit more on your ideas and the technical reasons behind this PR because I do have my clear doubts here.

sandmaennchen5:
The remember_ingress_credentials option came as an additional idea because I have seen similar optional functionality in other apps.

The main reason for these changes was that I had to log in again in every app whenever I switched between devices.

With remember_ingress_users: true, the OpenCCU SID is stored per Home Assistant user. This means that the same OpenCCU session can be reused across different devices as long as the SID is still valid and OpenCCU has not been restarted.

With ingress_keepalive_interval: 250, the stored OpenCCU sessions are periodically kept alive to prevent them from reaching the OpenCCU session timeout. The keepalive could also be disabled with 0.

So, theoretically, using only:
remember_ingress_users: true
ingress_keepalive_interval: 250

would already be enough to keep an OpenCCU session active and reusable across devices for each individual Home Assistant user.

remember_ingress_credentials adds another layer on top of that. When enabled, the OpenCCU login credentials are stored separately for each Home Assistant user and can be used to automatically log in again if the remembered SID is no longer valid or there is no active OpenCCU session.

The per-user data is stored under:
/mnt/data/supervisor/apps/data/a79d49ae_openccu-dev/etc/config/ha-ingress-sessions/SHA256(x-remote-user-id).json

Depending on the configuration, this file contains either only the last valid SID or additionally the encrypted OpenCCU username/password credentials.

Originally, the implementation with only remember_ingress_users and ingress_keepalive_interval required fewer changes. I added ingress_keepalive_interval because the maximum OpenCCU session timeout of 600 seconds was a bit too short for my use case.

In short:
- The already existing implementation basically remembers the SID only for the current HA Ingress/browser session and only for as long as that SID remains valid.
- remember_ingress_users remembers the SID per Home Assistant user, allowing all devices/sessions of that HA user to reuse it while it remains valid.
- ingress_keepalive_interval periodically keeps these stored OpenCCU sessions alive, potentially until OpenCCU is restarted.
- remember_ingress_credentials additionally stores the OpenCCU login credentials separately for each Home Assistant user and uses them to automatically log in again whenever a new OpenCCU session is required.

jens-maus:
@sandmaennchen5 Any reason why you closed this PR?

PR #4204, opened 2026-09-19
add multi-color LED support for RPI-RF-MOD

jens-maus:
This is a first bunch of changes to change the current single-color LED management to use the kernel own multicolor LED support which exists since kernel v6 and allow to group together three LEDs (RGB) to a single one. So instead of having to enable/disable and trigger each separate LED on a RPI-RF-MOD this kernel-driven multi-color LED support will allow to have a single sysfs LED node allowing to setup all different color combinations and especially trigger blink codes more synchronously. If used this should help to get blinking LED codes better synchronoused which in some cases currently are fading away from synchronous or alternating blinking resulting in some unwanted blinking sequences or strange color codes.

## Summary by CodeRabbit

* **New Features**
  * Added multicolor LED support and RGB status grouping for compatible boards.
  * Introduced centralized LED control with named slow and fast blink modes.
  * Added LED-only recovery builds and automatic LED service lifecycle handling.
* **Bug Fixes**
  * Improved driver ownership tracking to prevent unloading externally managed modules.
  * Updated status indicators across normal and recovery workflows.
* **Documentation**
  * Expanded LED control, platform, command-line, and recovery documentation.
* **Tests**
  * Added coverage for LED lifecycle, recovery behavior, help, and version output.


jens-maus:
@coderabbitai resume

